PR #164, opened 2026-07-01
fix: close five P1 enforcement-bypass issues (#159–#163)

### LINKED ISSUE #159: [P1] CONTEXT.md can be rewritten to disable every enforcement hook
Severity: P1 critical

## Problem

The activation bit that every enforcement hook trusts is ordinary repo content: `.codearbiter/CONTEXT.md` frontmatter with `arbiter: enabled`.

Source evidence:

- `plugins/ca/hooks/_hooklib.py:174-179` makes `arbiter_active(root)` return true only when `CONTEXT.md` has `arbiter: enabled`.
- `plugins/ca/hooks/pre-bash.py:332-336`, `pre-write.py:20-25`, and `pre-edit.py:21-25` all exit 0 immediately when `arbiter_active(root)` is false.
- `plugins/ca/hooks/pre-write.py:27-42` only blocks audit-log overwrites and ADR writes. It does not protect `.codearbiter/CONTEXT.md`.

That means the model can first rewrite `CONTEXT.md` to remove or change the flag, then every later hook goes dormant by design.

## Reproduction

In a throwaway arbiter-enabled repo on `main`:

1. Run `pre-write.py` with `file_path` set to `.codearbiter/CONTEXT.md` and content changing `arbiter: enabled` to `arbiter: disabled`.
2. The hook exits 0.
3. Before the actual file change, `pre-bash.py` blocks `git commit -m x` on main with H-01.
4. After writing the disabled frontmatter, the same `pre-bash.py` invocation exits 0.

Observed locally:

```text
pre-write CONTEXT overwrite exit 0
pre-bash before actual disable exit 2
pre-bash after disable exit 0
```

## Impact

An agent can disable the product's enforcement layer from inside the repo it is supposed to govern. Once dormant, direct commits, push guards, audit-log integrity, ADR authoring restrictions, security/migration backstops, and post-write reminders are all bypassed.

This contradicts the README's hard-gate claims because the opt-in activation file is also an unprotected kill switch.

## Smallest credible fix

Protect activation-critical project state:

- Block Write/Edit/MultiEdit and shell writes to `.codearbiter/CONTEXT.md` unless a sanctioned `/ca:init` or explicit reconfiguration marker is active.
- At minimum, block edits that remove or alter `arbiter: enabled` or corrupt the opening frontmatter.
- Add regression tests proving `CONTEXT.md` cannot be used to turn an enabled repo dormant before a guarded action.

## Duplicate check

Checked existing issues with:

```text
gh issue list --repo arbiterForge/codeArbiter --state all --search "CONTEXT.md arbiter enabled disable hooks activation bypass dormant"
```

No matching issue was returned.

### LINKED ISSUE #160: [P1] Gate pass markers are forgeable, so hard gates can self-approve
Severity: P1 critical

## Problem

Hard gates trust writable files under `.codearbiter/.markers/` as proof that a privileged review path already passed.

Source evidence:

- `plugins/ca/hooks/pre-bash.py:449-467` trusts `.codearbiter/.markers/security-gate-passed` when it is fresh and contains the digest of each sensitive added line.
- `plugins/ca/hooks/security-pass.py:75-90` writes that marker unconditionally for the current diff. The script itself does not verify that crypto-compliance or secret-handling actually ran.
- `plugins/ca/hooks/pre-bash.py:508-524` trusts `.codearbiter/.markers/migration-gate-passed` for migration commits.
- `plugins/ca/hooks/migration-pass.py:82-95` writes migration digests unconditionally for current migration files.
- `plugins/ca/skills/decision-lifecycle/SKILL.md:31-38` authorizes ADR writes by touching `.codearbiter/.markers/adr-authoring-active`.
- `plugins/ca/hooks/pre-write.py:36-42` and `pre-edit.py:64-70` allow ADR writes whenever that marker exists and is fresh.

The Write/Edit guards do not protect `.codearbiter/.markers/*`, and `pre-bash.py` does not block direct execution of `security-pass.py`, `migration-pass.py`, or `touch .codearbiter/.markers/adr-authoring-active`.

## Reproduction

In a throwaway enabled repo, stage a crypto change containing `hashlib.md5(...)`.

Control case:

```text
without marker exit 2
without marker stderr BLOCKED [H-09b]: This commit introduces crypto/TLS changes, but no security-gate pass is recorded ...
```

Then hand-write `.codearbiter/.markers/security-gate-passed` with the `line_digest()` of that added line, without running any security review.

Observed:

```text
forged marker exit 0
forged marker stderr
```

The same class applies to `migration-gate-passed` by writing `content_digest()` values, and to ADR authoring by creating `adr-authoring-active`.

## Impact

The marker files are the mechanism that turns a BLOCK into an allow. Because they are ordinary writable repo files and their producer scripts can be invoked directly, an agent can approve its own security, secret, migration, or ADR gate without the corresponding reviewer/skill ever running.

This collapses the hard-gate model to prose compliance.

## Smallest credible fix

Do not treat a repo-writable marker as sufficient authority by itself.

Possible directions:

- Block direct Write/Edit/Bash mutation of load-bearing marker files.
- Block direct `security-pass.py` and `migration-pass.py` execution outside a sanctioned gate context.
- Bind marker creation to a command-scoped nonce or signed receipt emitted by the reviewing gate, not just to diff content.
- Add regression tests that direct marker writes and direct marker-script invocations cannot unblock H-09b/H-10b/H-14/H-11.

## Duplicate check

Checked existing issues with:

```text
gh issue list --repo arbiterForge/codeArbiter --state all --search "security-gate-passed OR migration-gate-passed OR adr-authoring-active OR marker forge OR forge marker OR marker bypass"
```

No matching marker-forgery issue was returned.

### LINKED ISSUE #161: [P1] Shell indirection bypasses pre-bash commit, push, staging, and security gates
Severity: P1 critical

## Problem

`pre-bash.py` gates dangerous git operations by regex-matching the literal Bash/PowerShell command string. It does not evaluate shell indirection.

Source evidence:

- `plugins/ca/hooks/pre-bash.py:29-35` defines regexes for literal `git ... commit`, `git ... push`, and `git ... add` spellings.
- `plugins/ca/hooks/pre-bash.py:339-352` only enforces H-01 direct-commit blocking if `COMMIT_RE.search(cmd)` matched.
- `plugins/ca/hooks/pre-bash.py:413-467` only runs the H-09b/H-10b security scan inside the same `if commit:` branch.
- `plugins/ca/hooks/pre-bash.py:469-524` only runs the H-14 migration scan inside `if commit:` as well.

A command that constructs `git commit` through shell variables is still a real commit, but the hook sees no `git commit` token sequence and exits 0.

## Reproduction

In a throwaway arbiter-enabled repo on `main`:

```text
git commit -m x
exit 2
stderr BLOCKED [H-01]: Direct commit to main is prohibited (ORCHESTRATOR §3). Create a feature branch.

g=git; c=commit; $g $c -m x
exit 0
stderr
```

The second command would execute `git commit -m x` after the hook allows it.

## Impact

This bypasses the product's load-bearing shell hook for:

- direct commits to `main`/`master`
- protected-branch pushes and force pushes
- wildcard staging restrictions
- H-09b/H-10b crypto/secret commit backstops
- H-14 migration commit backstop

The bypass does not require exotic shell features; variable assignment and expansion are normal Bash/PowerShell usage.

## Smallest credible fix

Regex over a shell command string is not a reliable enforcement boundary for git mutations.

Possible directions:

- Install repo-level `pre-commit` / `pre-push` hooks for the hard backstops, so enforcement triggers on the git operation rather than the model's spelling.
- Fail closed on complex shell constructs in enabled repos when the command contains assignment/expansion/control operators near a git operation.
- Route commits/pushes through a sanctioned command path and block arbitrary Bash git mutation commands unless explicitly classified safe.

Add regression tests for at least variable indirection (`g=git; c=commit; $g $c ...`) and command-substitution/alias forms if supported by the shell surface.

## Duplicate check

Checked existing issues with:

```text
gh issue list --repo arbiterForge/codeArbiter --state all --search "pre-bash shell variable git commit bypass OR lexical bypass OR H-09b bypass OR H-01 bypass"
gh issue list --repo arbiterForge/codeArbiter --state all --search "security gate bypass commit hook pre-bash git variable"
```

No matching issue was returned.

### LINKED ISSUE #162: [P1] Symlink aliases bypass audit-log and ADR Write/Edit guards
Severity: P1 critical

## Problem

The Write/Edit guards decide whether a file is protected by matching the raw `file_path` string. They do not resolve symlinks before checking whether the target is under `.codearbiter/`.

Source evidence:

- `plugins/ca/hooks/pre-write.py:25` reads `tool_input.file_path` and normalizes separators only.
- `plugins/ca/hooks/pre-write.py:29-42` blocks only when `is_audit_log(fpath)` or `is_decisions_path(fpath)` matches that raw path.
- `plugins/ca/hooks/pre-edit.py:29-34` and `pre-edit.py:64-70` do the same raw-path checks.
- `_hooklib.repo_rel()` already resolves real paths in `plugins/ca/hooks/_hooklib.py:216-231`, but pre-write/pre-edit do not use it.

A symlink whose visible path does not contain `.codearbiter/` can point directly at `.codearbiter`, and the Write/Edit tools will follow the symlink after the hook allows the operation.

## Reproduction

In a throwaway enabled repo:

1. Create `.codearbiter/overrides.log` containing `seed`.
2. Create a symlink `alias -> .codearbiter`.
3. Invoke `pre-write.py` with `file_path` set to `<repo>/alias/overrides.log`.
4. Invoke `pre-edit.py` with `file_path` set to the same symlinked path.

Observed locally:

```text
pre-write.py <repo>/alias/overrides.log
exit 0
stderr

pre-edit.py <repo>/alias/overrides.log
exit 0
stderr

pre-write.py <repo>/.codearbiter/overrides.log
exit 2
stderr BLOCKED [H-05]: The .codearbiter audit logs ... are append-only ...
```

The canonical path blocks, while the symlink alias to the same target is allowed.

## Impact

An agent can overwrite or edit append-only audit logs and ADR files through a symlink alias, bypassing H-05 and H-11. This breaks the audit-trail and immutable-decision guarantees after a single unguarded symlink setup step.

## Smallest credible fix

Resolve protected file targets before classification:

- In pre-write/pre-edit, compute the repo-relative real path with the same `repo_rel()`/`realpath` approach used elsewhere.
- Apply H-05/H-11 to both the raw path and the resolved repo-relative target.
- Add tests for a symlinked directory and a symlinked file targeting `.codearbiter/overrides.log` and `.codearbiter/decisions/*.md`.
- Consider blocking symlink creation that targets protected `.codearbiter` paths through the Bash guard as defense in depth.

## Duplicate check

Checked existing issues with:

```text
gh issue list --repo arbiterForge/codeArbiter --state all --search "symlink audit log overrides.log pre-write pre-edit ADR decisions bypass"
```

No matching issue was returned.

### LINKED ISSUE #163: [P1] farm can recursively delete outside its workspace via FARM_WORKTREE_ROOT plus task id
Severity: P1 critical data loss

## Problem

The farm dispatcher resolves task worktree paths from an environment-controlled root and a plan-controlled task id, then recursively deletes that path before Git validates whether it is a real farm worktree.

Source evidence:

- `plugins/ca/tools/farm.ts:105` sets `ENV.worktreeRoot` from `FARM_WORKTREE_ROOT`, defaulting to `.farm/worktrees`.
- `plugins/ca/tools/farm.ts:1325` defines `SAFE_TASK_ID = /^[A-Za-z0-9._-]{1,64}$/`.
- `plugins/ca/tools/farm.ts:1378-1379` validates task ids only against that regex.
- `plugins/ca/tools/farm.ts:1069-1070` derives `wt = path.resolve(ENV.worktreeRoot, t.id)`.
- `plugins/ca/tools/farm.ts:885-891` calls `rm(wt, { recursive: true, force: true })` before `git worktree add`.

There is no check that the resolved `wt` is inside the repo's intended `.farm/worktrees` directory, nor that `FARM_WORKTREE_ROOT` is safe to delete beneath. The task id is plan-controlled and can name any child of the configured root that matches the regex.

## Reproduction shape

Do not run this against real data.

If an operator has `FARM_WORKTREE_ROOT=/Users/alice` and a plan contains a task id `Desktop`, the dispatcher computes:

```text
path.resolve('/Users/alice', 'Desktop') == '/Users/alice/Desktop'
```

Then `prepareWorktree()` reaches:

```text
rm('/Users/alice/Desktop', { recursive: true, force: true })
```

before Git can reject or create anything.

A less dramatic variant exists with `FARM_WORKTREE_ROOT=/tmp` and a task id matching an unrelated important directory under `/tmp`.

## Impact

A malformed or malicious plan plus a broad/misconfigured `FARM_WORKTREE_ROOT` can delete arbitrary directories under that root. The delete is recursive and forced. This is local data loss on the developer machine running the plugin.

The default root limits normal blast radius, but the env knob is documented configuration and the code does not enforce containment at the destructive operation.

## Smallest credible fix

Make deletion containment explicit and fail closed:

- Resolve the repo root and the allowed farm root once, then require every task worktree path to be inside that allowed root.
- Reject absolute or outside-repo `FARM_WORKTREE_ROOT` unless an explicit unsafe override is added.
- Reject task ids `.` and `..`, and consider using the stricter schema id pattern at runtime.
- Before any `rm(..., recursive: true)`, assert `isInside(allowedFarmRoot, wt)` and `wt !== allowedFarmRoot`.
- Add unit tests for `FARM_WORKTREE_ROOT=/Users/alice` plus task id `Desktop`, `.` and `..` ids, and a normal `.farm/worktrees/task-a` path.

## Duplicate check

Checked existing issues with:

```text
gh issue list --repo arbiterForge/codeArbiter --state all --search "FARM_WORKTREE_ROOT OR worktreeRoot OR SAFE_TASK_ID OR task id path traversal OR rm worktree"
gh issue list --repo arbiterForge/codeArbiter --state all --search "farm data loss OR worktree remove OR rm -rf OR path traversal"
```

No matching issue was returned.

SUaDtL:
## What & why

Five P1 reports (thanks @tg12) each showed the enforcement layer being enforced against a *forgeable or lexical* surface rather than the real underlying operation. Each is fixed below, and for every fix the rest of the repo was swept for the same class.

**#159 — CONTEXT.md was an unprotected kill switch.** Every hook gates on `arbiter: enabled` in `.codearbiter/CONTEXT.md`, but nothing protected that file — writing `arbiter: disabled` made every gate dormant. `pre-write`/`pre-edit` now block a CONTEXT.md write/edit whose *resulting* content drops `arbiter: enabled` or corrupts the frontmatter (`H-18`, reusing a new `frontmatter_enabled_text`); `pre-bash` adds the shell flank. Stage bumps that keep it enabled still pass.

**#160 — gate-pass markers were forgeable.** `.codearbiter/.markers/*` were writable repo files, so a hand-written marker forged a security/migration/ADR gate pass. `pre-write`/`pre-edit` (`H-19`) block tool writes to any marker; `pre-bash` blocks shell redirects/verbs naming `security-gate-passed`/`migration-gate-passed`. This closes every forge-*by-write* vector (the demonstrated exploit). Note: *blocking the producer scripts themselves is unworkable* — the sanctioned gate runs the identical command, so the hook can't distinguish a genuine post-review run from a forged one. That residual is inherent to a single-agent session and is documented in-code; the `/adr` `touch` and the python producers keep working.

**#161 — shell indirection bypassed the git gates → real git hooks.** `pre-bash` matches the literal command string, so `g=git; c=commit; $g $c` builds a real commit it never sees, and there was no enforcement below it. Added repo-level `pre-commit`/`pre-push` hooks (`git-enforce.py` installed by `_githooks.py`) that enforce H-01/H-02/H-09b/H-10b/H-14 at the git operation itself — spelling-proof. They reuse the same `_hooklib` detection primitives (no drift), fail closed on git read errors, select one interpreter (never `python3 X || python X`, which would swallow a BLOCK / drain pre-push stdin), honor `core.hooksPath`/worktrees, and **never clobber a foreign hook**. Installed idempotently at `/ca:init` and every SessionStart. `pre-bash` stays as the fast first line.

**#162 — symlink aliases bypassed the Write/Edit path guards.** `pre-write`/`pre-edit` classified the raw `file_path`, so a symlink whose visible path lacked `.codearbiter/` slipped past H-05/H-11. All protected-path checks now route through a new `classify_protected()` that resolves realpath (the approach `post-write-edit`/`pre-read` already used) and checks raw + resolved. Also wired `NotebookEdit` into the guard (it was matched by no hook).

**#163 — farm could `rm -rf` outside its workspace.** `prepareWorktree` resolved a delete path from env-controlled `FARM_WORKTREE_ROOT` + plan-controlled task id with no containment (and `SAFE_TASK_ID` admitted `.`/`..`). Now: an out-of-repo root is refused unless `FARM_ALLOW_EXTERNAL_WORKTREE_ROOT=1`; every worktree path is asserted strictly inside that root before the recursive delete (reusing the existing `isInside()`); and `validate()` rejects `.`/`..` ids.

Closes #159. Closes #160. Closes #161. Closes #162. Closes #163.

## Type of change

- [x] `fix`: bug fix
- [ ] `feat`: new behavior
- [ ] `docs`: documentation only
- [ ] `refactor`: no behavior change
- [ ] `chore`: deps, tooling, reverts

## Checklist

- [x] Branched off `main` (not a direct write to `main`)
- [x] Conventional Commits used in the commit messages
- [x] Tests added/updated for behavioral changes (`plugins/ca/hooks/tests/`)
- [x] Full suite green locally (585 hook tests + 5 CI guard matrices; farm typecheck + 180 vitest tests; `farm.js` rebuilt in sync)
- [ ] Version bumped if any shipped payload file changed — the current version `2.6.1` is **unpublished** (no `v2.6.1` tag), so these ride it and the CI version-bump gate passes; no bump needed
- [ ] `CHANGELOG.md` entry added — deferred to the per-release "roll the notes" step, matching repo convention
- [ ] ADR recorded — see reviewer notes: the git-hooks-install surface may warrant one
- [x] New behavior ships off-by-default / `preview` where appropriate — all new gates run **only in arbiter-enabled repos**; the git-hook install is idempotent and non-clobbering
- [ ] `docs/hooks.md` updated — see reviewer notes

## Notes for the reviewer

- **New hook files:** `git-enforce.py` (git-level enforcer), `_githooks.py` (installer). No network access added — `git-enforce.py` shells out only to local `git`, preserving the no-network invariant. `docs/hooks.md` was **not** updated in this PR; happy to follow up, or fold it in here.
- **The #160 scope call is deliberate** — I did not add a producer-blocking gate because it would break the legitimate gate flow (identical command line); the honest boundary is documented in `pre-bash.py`/`security-pass.py` comments. If you'd prefer the heavier SubagentStop-attestation route, that's a larger follow-up.
- **git-hook install surface** (`_githooks.py` wired into `init-codearbiter.py` + `session-start.py`) is the one arguably-architectural addition — flagging in case you want an ADR before merge.
- Tested cross-platform-safely (stdlib only, interpreter-fallback shim mirroring `hooks.json`), but the Windows `sh`-shim path was reasoned about, not run here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JqTxpxEkbDHb6AuADohtMR)_

### DIFF
--- a/.github/scripts/test_hooklib.py
+++ b/.github/scripts/test_hooklib.py
@@ -494,5 +494,44 @@ class MarkerDigestTest(unittest.TestCase):
         self.assertNotEqual(_hooklib.content_digest("a\nb\n"), _hooklib.content_digest("a\nc\n"))
 
 
+class ActivationAndMarkerHelpersTest(unittest.TestCase):
+    """#159/#160/#162 path classifiers and the text-based frontmatter check."""
+
+    def test_is_context_md(self):
+        self.assertTrue(_hooklib.is_context_md(".codearbiter/CONTEXT.md"))
+        self.assertTrue(_hooklib.is_context_md("/a/b/.codearbiter/CONTEXT.md"))
+        self.assertTrue(_hooklib.is_context_md(r".codearbiter\CONTEXT.md"))
+        self.assertFalse(_hooklib.is_context_md(".codearbiter/other.md"))
+        self.assertFalse(_hooklib.is_context_md("CONTEXT.md"))
+
+    def test_is_marker_path(self):
+        self.assertTrue(_hooklib.is_marker_path(".codearbiter/.markers/security-gate-passed"))
+        self.assertTrue(_hooklib.is_marker_path("x/.codearbiter/.markers/adr-authoring-active"))
+        self.assertFalse(_hooklib.is_marker_path(".codearbiter/CONTEXT.md"))
+        self.assertFalse(_hooklib.is_marker_path(".codearbiter/markers/x"))  # no dot
+
+    def test_frontmatter_enabled_text(self):
+        self.assertEqual(_hooklib.frontmatter_enabled_text("---\narbiter: enabled\n---\n"),
+                         (True, False))
+        self.assertEqual(_hooklib.frontmatter_enabled_text("---\narbiter: disabled\n---\n"),
+                         (False, False))
+        # opened, never closed -> malformed
+        self.assertEqual(_hooklib.frontmatter_enabled_text("---\narbiter: enabled\n"),
+                         (False, True))
+        # no frontmatter at all -> dormant, not malformed
+        self.assertEqual(_hooklib.frontmatter_enabled_text("# ctx\n"), (False, False))
+
+    def test_classify_protected_resolves_symlink(self):
+        with tempfile.TemporaryDirectory() as d:
+            ca = os.path.join(d, ".codearbiter")
+            os.makedirs(ca)
+            with open(os.path.join(ca, "overrides.log"), "w") as f:
+                f.write("x")
+            os.symlink(ca, os.path.join(d, "alias"))
+            # raw path lacks .codearbiter/, but realpath resolves inside it.
+            hits = _hooklib.classify_protected(os.path.join(d, "alias", "overrides.log"), d)
+            self.assertIn("audit", hits)
+
+
 if __name__ == "__main__":
     unittest.main(verbosity=2)
--- a/.github/scripts/test_hooklib.py
+++ b/.github/scripts/test_hooklib.py
@@ -26,6 +26,17 @@ sys.path.insert(0, HOOKS)
 import _hooklib  # noqa: E402 — needs sys.path mutation above
 
 
+def _sym_ok():
+    """Windows CI runners often lack symlink privilege; skip symlink-dependent
+    cases there (ubuntu/macos exercise them fully)."""
+    try:
+        with tempfile.TemporaryDirectory() as d:
+            os.symlink(os.path.join(d, "t"), os.path.join(d, "l"))
+        return True
+    except (OSError, NotImplementedError, AttributeError):
+        return False
+
+
 class CryptoReTest(unittest.TestCase):
     """CRYPTO_RE drives the H-09/H-09b crypto gate. It must see the TLS-disable
     and banned-primitive forms in BOTH Python and Node/TS, since all networked
@@ -521,6 +532,7 @@ class ActivationAndMarkerHelpersTest(unittest.TestCase):
         # no frontmatter at all -> dormant, not malformed
         self.assertEqual(_hooklib.frontmatter_enabled_text("# ctx\n"), (False, False))
 
+    @unittest.skipUnless(_sym_ok(), "symlink creation not permitted here")
     def test_classify_protected_resolves_symlink(self):
         with tempfile.TemporaryDirectory() as d:
             ca = os.path.join(d, ".codearbiter")
--- a/plugins/ca/hooks/_githooks.py
+++ b/plugins/ca/hooks/_githooks.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+# codeArbiter — installs the git-level enforcement hooks (#161).
+#
+# The PreToolUse Bash hook (pre-bash.py) gates git operations by matching the
+# literal command string, so shell indirection (`g=git; c=commit; $g $c`) walks
+# past it. There is no enforcement below that layer. This module installs
+# repo-level .git/hooks/pre-commit and pre-push that invoke git-enforce.py at the
+# git operation itself, where spelling no longer matters.
+#
+# Design decisions:
+#   * The shim is a tiny POSIX `sh` script that detects the interpreter ONCE
+#     (python3 else python) and runs the enforcer EXACTLY once — never
+#     `python3 X || python X`, which would (a) swallow a BLOCK when python3 both
+#     exists and blocks, and (b) drain stdin before the fallback (pre-push feeds
+#     the ref list on stdin). Same hazard hooks.json avoids via two entries; a
+#     single hook file must guard it inline.
+#   * The shim points at the enforcer by ABSOLUTE PATH, resolved from THIS file's
+#     location (inside the plugin) at install time. install() is re-run every
+#     SessionStart, so the path is refreshed each session — if a plugin update
+#     moves the install dir, the next session rewrites the shim. During the brief
+#     window before that, a missing enforcer makes the shim exit 0 (fail-OPEN on
+#     our OWN staleness only — never brick a user's commits because our path
+#     drifted; the pre-bash + Claude layers still apply).
+#   * A pre-existing NON-ours hook is NEVER clobbered — we warn loudly and skip,
+#     so an existing husky / pre-commit-framework setup is preserved.
+#   * Idempotent: an up-to-date ours-hook is left untouched (no churn); a stale
+#     ours-hook is refreshed.
+
+import os
+import stat
+import subprocess
+import sys
+
+SENTINEL = "# codeArbiter-managed git hook (#161) — refreshed each session; edits are overwritten."
+PHASES = ("pre-commit", "pre-push")
+
+
+def _warn(msg):
+    print(f"codeArbiter git-hooks: {msg}", file=sys.stderr)
+
+
+def _git(args, cwd):
+    try:
+        return subprocess.run(
+            ["git"] + args, cwd=cwd, capture_output=True, text=True,
+            encoding="utf-8", errors="replace", timeout=5,
+        )
+    except Exception:  # noqa: BLE001
+        return None
+
+
+def hooks_dir(root):
+    """The directory git actually reads hooks from for `root`, or None.
+
+    Honors core.hooksPath (when set, git IGNORES .git/hooks entirely), and
+    resolves the real git dir via `rev-parse --git-path hooks` so linked
+    worktrees and submodules land in the right place. Falls back to
+    <root>/.git/hooks only if git can't answer."""
+    cfg = _git(["config", "--get", "core.hooksPath"], root)
+    if cfg is not None and cfg.returncode == 0 and cfg.stdout.strip():
+        hp = cfg.stdout.strip()
+        return hp if os.path.isabs(hp) else os.path.join(root, hp)
+    gp = _git(["rev-parse", "--git-path", "hooks"], root)
+    if gp is not None and gp.returncode == 0 and gp.stdout.strip():
+        hp = gp.stdout.strip()
+        return hp if os.path.isabs(hp) else os.path.join(root, hp)
+    default = os.path.join(root, ".git", "hooks")
+    return default if os.path.isdir(os.path.join(root, ".git")) else None
+
+
+def _enforcer_path():
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "git-enforce.py")
+
+
+def _shim(enforcer, phase):
+    # Single-interpreter selection preserves stdin (pre-push) and the BLOCK exit
+    # code. `exit 0` when the enforcer file is absent is deliberate fail-open on
+    # our own path staleness (see module header).
+    return (
+        "#!/bin/sh\n"
+        f"{SENTINEL}\n"
+        f'E="{enforcer}"\n'
+        '[ -f "$E" ] || exit 0\n'
+        'if python3 -c "" 2>/dev/null; then PY=python3; else PY=python; fi\n'
+        f'exec "$PY" "$E" {phase}\n'
+    )
+
+
+def _read(path):
+    try:
+        with open(path, encoding="utf-8", errors="replace") as f:
+            return f.read()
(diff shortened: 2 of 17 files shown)


SUaDtL:
Thanks @tg12. these five P1 reports were exceptionally well-researched: precise source evidence, reproductions, and correct duplicate checks on every one. Each validated exactly as written, and #164 fixes all of them (#159–#163), with the rest of the repo swept for the same class in each case. Much appreciated.

Issue #259, opened 2026-07-09
ca-codex ships ~15 unreachable entry scripts and a broken /ca-init first-run pointer


SUaDtL:
# ca-codex ships ~15 unreachable entry scripts and a broken /ca-init first-run pointer

> **Codex multi-host support (ADR-0011).** Code under review is on branch `feat/codex-support-m0`, not `main`. ca-codex ships BETA; these are blocking-severity for the affected code but nothing is merged to the default branch yet.

**Severity:** med  |  **Confidence:** 0.8  |  **Group:** dead-vendored-entries

**Where:**
- `plugins/ca-codex/hooks/hooks.json:1-70`
- `plugins/ca-codex/.codex-plugin/plugin.json:5`
- `tools/sync-core.py:41-50`

**Evidence / impact:**
- (architecture-005) hooks.json registers exactly 5 entries (session-start, pre-bash, pre-write, post-write-edit, prune-transcript). The plugin ships no commands/, skills/, or agents/ (`ls plugins/ca-codex/` = hooks + ORCHESTRATOR.md + manifest). Yet sync-core.py vendors ALL 42 core .py files, so statusline.py, wire-statusline.py, pre-read.py, pre-edit.py, git-enforce.py, doctor.py, preview.py, metrics.py, taskwrite.p
- (architecture-005 impact) Dead code masquerading as live: two-thirds of the shipped Codex payload is unreachable, inflating the trust-review surface users must approve and blurring which files are enforcement-bearing. The /ca-init pointer is a broken first-run path for the shipped beta: an installing user cannot follow the m

**Recommendation:**

Capability-gate or trim the vendored payload so ca-codex does not ship ~15 non-registered enforcement scripts as trust-surface bloat, and fix the manifest first-run pointer (it names a /ca-init that ca-codex does not ship).

**Acceptance criteria:**
- ca-codex's shipped payload contains only files reachable on that host, or unreachable ones are clearly marked.
- The manifest's documented first-run command exists in the shipped plugin.

SUaDtL:
Partially addressed by #285: the broken `/ca-init` manifest pointer is fixed (accurate activation description). 

Re-scoping the remaining two concerns:
[...]

PR #304, opened 2026-07-13
feat: clear auto-safe open issues

### LINKED ISSUE #278: flaky: test_prune_nudge warm-reset test is wall-clock dependent (false reds on Windows CI)
`test_cold_arm_then_warm_clears_and_persists_through_prune` (plugins/ca/hooks/tests/test_prune_nudge.py:592) stamps a "warm" transcript timestamp via `_warm_ts()` and expects `hook_run` to still classify it as warm when it evaluates it. On a slow/loaded runner the wall-clock gap between stamping and evaluation can push the timestamp past the idle/warm threshold, so it reads as cold, `cold_nudged` is not cleared, and the test fails with `AssertionError: True is not false : warm submit must persist the cold_nudged clear`.

Observed: intermittent failure on GitHub `windows-latest` (CI run 29052450158); passes on rerun and passes locally 5x (standalone + under `unittest discover`). Pre-existing; not caused by any codex-support change.

Impact: false-red CI on unrelated PRs, which erodes "merge green" discipline.

Fix: make the test deterministic — inject/patch the clock (or the idle computation) so "warm" does not depend on real elapsed time between stamping the fixture and evaluating it. Audit sibling `_warm_ts()`/`_idle`-based tests for the same dependency.

### LINKED ISSUE #283: docs(site): gate-events.log format in codearbiter-directory.md is stale after host= attribution (#269)
After #269 (ADR-0012 host attribution), the gate-events.log line format is `[ISO-8601Z] KIND [tag] host=<host> hook=<script> | msg`, but `site/src/content/docs/codearbiter-directory.md:225` still documents the pre-host= form `[ISO-8601Z] KIND [tag] hook=<script> | msg`.

Trivial one-line doc update. Kept out of the #269 hooks PR because a `site/` change needs its own npm-test verification (vitest content-structure suite) and should not ride a hooks-only PR.

Fix: update line 225 to include `host=<host>`, then run `npm test` in `site/`.

### LINKED ISSUE #293: site generator hook-gate tests pin exact pre-bash.py line numbers (brittle; breaks late on main-PRs)
The real-hooks snapshot tests in `site/test/generator/extract-hook-gates.test.ts` pin gate call sites to exact line numbers in `plugins/ca/hooks/pre-bash.py`. ANY edit above those lines (even a comment) breaks the suite — and because the site vitest job only runs on main-targeted PRs, the break surfaces late (it blocked feat→main PR #254 after the codex remediation shifted lines +9; fixed by updating the pins). A structural key (tag + verbatim-message match, or tag + ordinal-within-file) would keep the verbatim-extraction guarantee without the line coupling.

### LINKED ISSUE #296: statusline ledger loses session updates under concurrent renders
## Problem

The user-global statusline ledger performs an unlocked read-modify-write transaction. Per-process temporary filenames prevent temporary-file corruption, but they do not prevent last-writer-wins replacement of a newer ledger snapshot.

Two Claude Code sessions rendering at the same time can both read the same prior ledger, update different session records, and then replace the ledger in sequence. The later replacement silently drops the earlier session's update. Daily totals and cached session state can consequently move backward or disappear.

## Evidence

- `core/pysrc/_ledgerlib.py:252` — `ledger_update()` reads and mutates the shared ledger.
- `core/pysrc/_ledgerlib.py:308-311` — a PID-scoped temporary file is atomically replaced into the shared path.
- `core/pysrc/_ledgerlib.py:330-359` — `persist_sess_start()` performs a second independent read-modify-write transaction against the same file.

Atomic replacement protects file integrity, not transaction isolation.

## Recommended direction

Prefer one independently written ledger file per session, aggregated during render. A cross-platform lock around shared-ledger transactions is also viable, but it adds platform-specific failure and stale-lock behavior to a latency-sensitive path.

## Acceptance criteria

- Concurrent updates for two distinct session IDs both survive.
- `persist_sess_start()` cannot discard a concurrent token/cost update.
- Interrupted writes never corrupt previously valid ledger data.
- Today/session totals remain correct after controlled concurrent execution.
- Tests exercise the collision deterministically rather than relying on timing luck.

### LINKED ISSUE #297: statusline reports no Git branch in linked worktrees
## Problem

The statusline's branch resolver assumes `.git` is a directory containing `.git/HEAD`. In a linked Git worktree, `.git` is a text file containing a `gitdir:` pointer. Opening `.git/HEAD` fails, `head_branch()` returns `None`, and the rendered header can incorrectly report `no git`.

This matters for codeArbiter's own worktree-based development workflows.

## Evidence

- `core/pysrc/_gitlib.py:47-59` reads `<root>/.git/HEAD` directly.
- `core/pysrc/statusline.py:526-535` uses that result to decide whether Git context exists.

## Recommended direction

Resolve a `gitdir:` pointer and read the actual HEAD, or add a short timeout-bounded `git rev-parse` fallback. Preserve the direct-file fast path for ordinary checkouts.

## Acceptance criteria

- Normal checkouts still show the current branch.
- Linked worktrees show their current branch.
- Detached HEAD continues to show a short commit ID.
- Missing or malformed Git metadata degrades without traceback.
- Tests include a realistic `.git` pointer-file fixture.

### LINKED ISSUE #298: statusline dirty check can add up to 1.5 seconds to every render
## Problem

Every statusline render with a detected branch runs a full `git status --porcelain` and permits it to block for up to 1.5 seconds. Large repositories, network filesystems, and expensive untracked-file scans can therefore make a UI-only statusline visibly lag.

## Evidence

- `core/pysrc/_gitlib.py:62-68` invokes `git status --porcelain` with `timeout=1.5`.
- `core/pysrc/statusline.py:533-535` performs the dirty check while constructing every header.

## Recommended direction

Benchmark representative repositories before selecting among:

1. a short-lived cache keyed by repository/worktree;
2. a tighter latency budget with fail-soft behavior;
3. a cheaper status mode, potentially excluding untracked files if that semantic tradeoff is documented.

## Acceptance criteria

- The statusline has a defined and tested dirty-check latency budget.
- Timeout or Git failure still degrades without traceback.
- Normal tracked changes continue to produce the dirty marker.
- Any change to untracked-file semantics is explicit and tested.
- A performance regression test or deterministic timeout fixture covers the slow-Git path.

### LINKED ISSUE #299: statusline ownership detection can mistake third-party commands for codeArbiter
## Problem

The wiring helper decides that a statusline belongs to codeArbiter when its command contains the generic substring `statusline.py`. A third-party statusline using the same filename can therefore be treated as codeArbiter-owned and refreshed in place without first being backed up.

That creates a configuration-loss risk during install or refresh.

## Evidence

- `core/pysrc/wire-statusline.py:40` defines `MARKER = statusline.py`.
- `core/pysrc/wire-statusline.py:98-103` uses substring membership as the ownership test.
- `core/pysrc/wire-statusline.py:161-166` refreshes anything recognized as owned.

## Recommended direction

Use a specific codeArbiter path signature or explicit ownership metadata. Keep compatibility with already-installed codeArbiter commands so upgrades remain seamless.

## Acceptance criteria

- A third-party command ending in `statusline.py` is not considered owned.
- Current and stale versioned codeArbiter statusline paths are recognized.
- Install backs up an unowned existing statusline.
- Refresh leaves an unowned existing statusline untouched.
- Uninstall restores the exact prior setting.

### LINKED ISSUE #300: feat(statusline): named themes and a validated custom palette
## Feature request

Let users change the Claude Code statusline palette without changing its information architecture, while preserving the existing neon-violet presentation as the default.

The feature should serve two audiences:

1. users who want a simple named alternative; and
2. users who want to supply their own palette without editing generated plugin files.

### Built-in themes

Select a curated theme with:

```text
CODEARBITER_THEME=violet
CODEARBITER_THEME=blue
CODEARBITER_THEME=green
CODEARBITER_THEME=amber
CODEARBITER_THEME=mono
```

- `violet` remains the default and must be byte-for-byte visually equivalent to the current palette.
- Theme names are case-insensitive.
- An unknown name fails softly to `violet`; it must never blank or break the statusline.
- `NO_COLOR` continues to take precedence and strips all SGR sequences regardless of theme.

### Custom palette

Support a persistent JSON palette file, defaulting to:

```text
~/.codearbiter/statusline-theme.json
```

A path override may be supplied with `CODEARBITER_THEME_FILE`. The custom file is activated with `CODEARBITER_THEME=custom`.

Suggested schema:

```json
{
  accent: {
    deep: #6c46b4,
    mid: #965ce6,
    primary: #b266ff,
    bright: #d08cff
  },
  text: {
    muted: #9696a2,
    normal: #e8e8f0,
    on_accent: #120e1a
  },
  semantic: {
    ok: #78dc96,
    warn: #ffb84c,
    danger: #ff566e
  },
  gradient: {
    from: #7850c8,
    to: #cd8cff
  }
}
```

The custom palette may be partial: omitted keys inherit from `violet`. Invalid keys or color values are ignored with fail-soft fallback. Rendering must remain dependency-free and must never print a traceback.

### 1. Make palette resolution explicit in `_colorlib.py`

Replace the module's hardcoded public color constants with a small immutable palette model and a resolver:

```python
@dataclass(frozen=True)
class Palette:
    accent_deep: RGB
    accent_mid: RGB
    accent_primary: RGB
    accent_bright: RGB
    text_muted: RGB
    text_normal: RGB
    text_on_accent: RGB
    semantic_ok: RGB
    semantic_warn: RGB
    semantic_danger: RGB
    gradient_from: RGB
    gradient_to: RGB

def resolve_palette(theme_name=None, custom_path=None) -> Palette:
    ...
```

Keep the existing exported names (`V0`, `V1`, `V2`, `V3`, `GREY`, `WHITE`, `OK`, `WARN`, `DANGER`, and `PILL_FG`) as compatibility projections from the resolved palette. This avoids a broad rewrite of `statusline.py`, `_boxlib.py`, `_fmtlib.py`, and `_segmentslib.py`.

### 2. Centralize built-in definitions

Store built-in themes as data in `_colorlib.py` or a dedicated dependency-free sibling such as `_themelib.py`. Do not scatter theme branches through individual segments.

Semantic colors must remain semantically stable: danger is still danger, warning is still warning, and success is still success. Built-ins should be checked for readable contrast rather than merely shifting every RGB channel.

### 3. Generalize gradients

Change `gradient_h()` so its default endpoints come from the active palette rather than hardcoded violet RGB tuples. Explicit `c_from` and `c_to` arguments must continue to work for compatibility.

### 4. Load custom data defensively

- Expand `~` and environment variables in the configured path.
- Bound the file size before parsing.
- Accept six-digit `#RRGGBB` values only for the first version.
- Ignore unknown keys.
- Merge a partial valid file over the violet defaults.
- On missing, unreadable, oversized, or malformed input, render with violet.
- Perform no writes and no network access during rendering.

### 5. Preserve generated-source discipline

Implement in `core/pysrc/`, then materialize the host payloads through `tools/sync-core.py`. Do not hand-edit vendored copies under plugin directories.

Although Codex currently has no statusline surface, keeping the shared generated payload byte-identical avoids host forks and leaves the palette logic ready if Codex gains such a surface later.

### 6. Document configuration

Add the built-in selector, custom schema, fallback behavior, and `NO_COLOR` precedence to the Claude statusline documentation. Include screenshots or terminal captures of every built-in theme, but do not imply Codex renders this statusline.

## Testing

Add focused tests for:

- violet default RGB values remain unchanged;
- every built-in resolves to a complete palette;
- theme names are case-insensitive;
- unknown names fall back to violet;
- a complete custom palette;
- partial custom palette inheritance;
- malformed JSON, wrong types, invalid hex, unknown keys, missing file, and oversized file;
- custom gradient endpoints;
- `NO_COLOR` strips themed output;
- ANSI-aware width, clipping, and padding remain unchanged under every theme;
- semantic threshold colors still switch at the existing context/rate boundaries;
- shared-source byte identity remains clean after generation.

## Acceptance criteria

- Existing users see no visual change unless they opt in.
- At least five built-in themes are available: violet, blue, green, amber, and monochrome.
- Users can provide a partial or complete custom palette without editing plugin files.
- Invalid theme configuration never produces a traceback or blank statusline.
- `NO_COLOR` remains authoritative.
- Theme selection changes color only; layout, glyphs, content, thresholds, and status semantics do not change.
- The renderer remains Python-stdlib-only and performs no network activity.
- Documentation clearly labels the statusline as Claude Code-only.

## Additional scope: show each subagent's selected model

The statusline currently shows each recent subagent's label, token totals, age, and liveness, but not the model that performed the work. Users should not have to infer whether a dispatcher selected Opus, Sonnet, Haiku, or another model.

The subagent transcript already exposes the model on assistant messages. Extend `read_subagents()` in `core/pysrc/_subagentslib.py` to capture it while performing the existing bounded JSONL scan?no second file pass and no new subprocess.

Illustrative parser change:

```python
def display_model(model_id):
    """Compact a host model ID without hiding its family/version."""
    raw = str(model_id or "").strip()
    if not raw:
        return "model:?"
    name = raw.removeprefix("claude-")
    name = re.sub(r"-(\d{8}|\d{4}-\d{2}-\d{2})$", "", name)
    return "model:" + name[:24]


# Inside read_subagents(), alongside label and reqs:
models = []

# Inside the existing parsed-message loop:
if msg.get("role") == "assistant" and msg.get("model"):
    model = str(msg["model"])
    if model not in models:
        models.append(model)

# When constructing the shown record:
shown.append({
    "label": label or fallback_label,
    "model": display_model(models[0]) if len(models) == 1 else (
        "model:mixed" if models else "model:?"
    ),
    "inp": inp,
    "out": out,
    "age": now - mtime,
    "active": now - mtime <= ACTIVE_WINDOW,
})
```

Then include the model in the existing right-hand metrics tail in `core/pysrc/statusline.py`, so label clipping automatically yields space to model information:

```python
model = f"{V3}{sub.get('model', 'model:?')}{RESET}"
tail = (
    f"  {model}"
    f" {V2}{DN}{RESET} {GREY}{fmt_tok(sub['inp'])}{RESET}"
    f" {V2}{UP}{RESET} {GREY}{fmt_tok(sub['out'])}{RESET}"
    f"  {DIM}{human_dur(sub['age'])}{RESET}"
)
```

The active theme should color the model tag through the normal accent vocabulary; model families must not introduce their own hardcoded colors.

Additional acceptance criteria:

- Every displayed subagent row includes its recorded model or an explicit `model:?` fallback.
- Known Claude model IDs retain enough family/version information to distinguish selections such as Opus, Sonnet, and Haiku.
- If a transcript contains assistant calls from multiple model IDs, the row says `model:mixed` rather than reporting a misleading single model.
- Model extraction happens during the existing bounded transcript scan.
- Narrow terminals clip the task label before dropping the model tag, token totals, or age.
- Malformed or absent model metadata never breaks the statusline.
- Tests cover one model, multiple models, absent metadata, long unknown model IDs, and narrow-width rendering.

### LINKED ISSUE #305: Windows: hooks invoking python3 via MSIX app-execution alias trigger unbounded AppXSvc memory leak
## Summary

On Windows, the `ca` plugin's hooks invoke `python3` on every hook firing (SessionStart, PreToolUse, etc. — see `hooks/hooks.json`, which additionally runs a `python3 -c "" || python ...` fallback, so up to two `python3` launches per hook entry). If the user's `python3` resolves to a Windows **app execution alias** for an MSIX-packaged Python — notably python.org's **Python Install Manager** (`PythonSoftwareFoundation.PythonManager`), which is distributed via `.appinstaller` — then *every hook execution* launches the package through the MSIX activation path, and Windows runs an `UpdateUsingAppInstallerOperation` update check inside **AppXSvc** on each launch.

AppXSvc leaks memory per operation and runs as a Protected Process Light (`LaunchProtected=2`), so it cannot be killed even by SYSTEM — the leak accumulates until reboot.

## Observed impact

- Machine: Windows 11 Pro (26200), heavy Claude Code + Codex usage with the `ca` / `ca-codex` plugins active.
- `Microsoft-Windows-AppXDeploymentServer/Operational` recorded **6,628 event-603 `UpdateUsingAppInstallerOperation` starts in ~25 hours** (roughly one every few seconds), all for `PythonSoftwareFoundation.PythonManager`. Bursts correlate 1:1 with plugin hook firings during active sessions.
- The svchost hosting AppXSvc grew to **25.5 GB private bytes** and could not be terminated (`Stop-Process`/`taskkill` fail with access denied even via PsExec `-s`, due to PPL).

## Reproduction sketch

1. Windows machine with the Python Install Manager installed (its `python3.exe` app-execution alias enabled — the default) and no earlier `python3` on PATH.
2. Install the `ca` plugin; run any Claude Code session so hooks fire repeatedly.
3. Watch `Get-WinEvent -LogName Microsoft-Windows-AppXDeploymentServer/Operational` — one or two 603 events per hook firing; AppXSvc private bytes climb and never recede.

## Possible plugin-level mitigations

- Resolve the interpreter **once** per session (e.g. in the SessionStart hook: probe for a non-alias interpreter — `sys.executable` of a successful probe run, or `py -3` on Windows — cache the absolute path, and have subsequent hooks use it) instead of re-resolving `python3` through PATH on every hook.
- On Windows, prefer `py`/`python` before `python3` (the alias trap is specific to `python3`, which classic python.org installers never provide — on Windows a bare `python3` is *usually* an MSIX alias).
- Detect the app-execution-alias case (the resolved `python3.exe` under `%LOCALAPPDATA%\Microsoft\WindowsApps\` is an `IO_REPARSE_TAG_APPEXECLINK` reparse point) and warn the user at SessionStart.
- Drop the unconditional `python3 -c "" || python ...` second hook entry where possible — it doubles the alias launches even when the first entry succeeded.

## If no code fix: document it

At minimum, the docs should warn Windows users that hooks call `python3` per tool-use, and that with the Python Install Manager's app-execution alias this triggers an AppXSvc update check per call, leading to unbounded AppXSvc memory growth. Workaround for users:

- Settings → Apps → Advanced app settings → App execution aliases → disable the Python Install Manager's `python3.exe` alias (or remove the reparse point at `%LOCALAPPDATA%\Microsoft\WindowsApps\python3.exe`), so `python3` resolves to `%LOCALAPPDATA%\Python\bin\python3.exe` directly, bypassing MSIX activation entirely.
- Reboot (or restart AppXSvc, if not PPL-blocked) to reclaim already-leaked memory; keep AppXSvc at its default `Manual` start type.

### LINKED ISSUE #306: ca-codex: pre-tool-adapter.py / pre-write.py hang blocking on stdin, orphaned processes accumulate
## Summary

`ca-codex` (observed on 0.2.4) hook processes — `hooks/pre-tool-adapter.py` and the `hooks/pre-write.py` it spawns — can hang forever blocking on stdin and accumulate as orphaned processes when the parent Codex session ends without feeding/closing their stdin.

## Observed impact

Over one workday of Codex usage, **16 orphaned `python.exe` processes** (8 adapter/pre-write pairs) accumulated, with start times spanning ~11:33 AM to ~1:56 PM and still alive at ~11:30 PM. Each pair is `C:\Python314\python.exe .../ca-codex/0.2.4/hooks/pre-tool-adapter.py` with a child `... hooks/pre-write.py`; the adapter's own parent is gone (orphaned). ~40–50 MB RSS each — not huge individually, but unbounded growth across sessions/days.

## Likely cause

The hook scripts do a blocking `sys.stdin.read()` (or equivalent) with no timeout and no handling for the case where the parent dies while the pipe is still open on the child side (the adapter holding the write end of pre-write's stdin keeps *both* alive, since the adapter itself is blocked).

## Suggested fix

- Read stdin with a timeout (e.g. `selectors`/threading watchdog on Windows, where `select()` doesn't work on pipes) and exit non-zero if no input arrives within a few seconds.
- Treat EOF/broken pipe as a terminal condition and exit immediately.
- Optionally: have the adapter watch its parent PID and self-terminate (and kill its pre-write child) when the parent disappears.

## Workaround for affected users

```powershell
Get-Process python -ErrorAction SilentlyContinue |
  Where-Object { (Get-CimInstance Win32_Process -Filter "ProcessId=$($_.Id)").CommandLine -like '*ca-codex*' } |
  Stop-Process -Force
```

SUaDtL:
## Summary

Clears the approved auto-safe issue set with deterministic regressions and generated-source parity across Claude and Codex payloads.

- Makes ledger persistence, gate-event appends, linked-worktree branch detection, ownership checks, and dirty-state rendering safe under concurrency and failure.
- Adds five opt-in statusline palettes, bounded custom JSON themes, and subagent model labels while preserving default violet output.
- Removes timing and source-line brittleness from prune and documentation tests, and documents gate-event host attribution.
- Records ca 2.9.0 and ca-codex 0.3.0 release metadata plus the completed sprint governance trail.

The gate-event fixes follow conflict hierarchy level 1: audit-trail correctness took priority after deterministic Windows failures blocked the sprint gate. The user explicitly approved that scope addition.

Closes #278
Closes #283
Closes #293
Closes #296
Closes #297
Closes #298
Closes #299
Closes #300

Issue #259 was validated against the shipped Codex bootstrap and payload contract, then closed with evidence.

## Test plan

- [x] Full hook suite: 930 tests
- [x] All repository commit-gate scripts
- [x] Real Windows dual-host concurrency suite: 2 tests
- [x] Site Vitest suite: 403 tests
- [x] Site typecheck, build, and link audit
- [x] Canonical sync: 42 files across 2 payloads, byte-identical
- [x] Build-surface and plugin-reference checks
- [x] Final specification, quality, and coverage reviews with no blocking findings

Closes #305
Closes #306


SUaDtL:
/ca:review verdict (round 2, 2026-07-13): PASS.

- Round 1 blocked on a HIGH (hooks.json python-first flip failed open on Python-2 hosts) and red POSIX CI. Both resolved by the version-aware fallback probe (a68389e7) with static + dynamic PY2 regression coverage; CI green on all three runners.
- Governance attributions added by the external agent (overrides.log H-05 override entry, sprint-log D-02 approval resolution) were confirmed genuine by the user on the record in the review session. No correction needed; entries stand as-is.
- Architecture drift: none (9 ADRs confirmed; core-to-payload parity byte-identical).
- Five deferrable findings (1 MEDIUM test gap, 4 LOW hardening items) harvested to .codearbiter/open-tasks.md with origin pr304-review.

No BLOCK-level findings stand. Merge decision rests with the user.

PR #347, opened 2026-07-20
fix(hooks): block Git rollback of audit logs

### LINKED ISSUE #335: H-05 gap: git checkout -- <audit-log> reverts append-only logs unblocked
Found live during sprint backlog-burn-2026-07-17 (user steer: file plugin oddities as issues). A subagent ran `git checkout -- .codearbiter/gate-events.log .codearbiter/sprint-log.md` to 'clean up' side-effect diffs and silently reverted uncommitted appends on two append-only audit artifacts. The H-05 flanks cover rm/redirect/truncate spellings and Write/Edit, but git-native restore spellings (`git checkout -- <path>`, `git restore <path>`, `git checkout HEAD -- <path>`) reach pre-bash unblocked while a filename token sits adjacent to the verb — inside the current lexical guard's intent, unlike the accepted $var-indirection residual class.

Remediation shape: extend pre-bash H-05 to block checkout/restore (and stash apply/pop touching audit paths?) when an AUDIT_LOG_NAMES basename appears in argv; tests in test_hook_guards.py for both spellings plus the -- form. sprint-log content was reconstructed with a visible AUDIT-CORRECTION entry; gate-events hook telemetry between last commit and revert is unrecoverable.

SUaDtL:
## Summary

- Block `git checkout` and `git restore` when the invocation names an append-only codeArbiter audit artifact.
- Parse separated-value Git global options such as `--git-dir .git` and `--work-tree .` through the same grammar used for command matching and worktree cwd extraction.
- Preserve read-only Git inspection and command-boundary isolation.

## Why

A live sprint incident showed that Git-native restore commands could erase uncommitted audit history without crossing H-05's filesystem-verb checks. The required coverage review then found that valid separated-value global options could hide the same restore operation. Both paths now fail closed.

Tradeoff: H-05 remains a lexical shell flank. It blocks literal audit-log names in checkout and restore invocations, but does not infer the contents of pathless stash operations or broad pathspecs. That preserves the existing Level 1 enforcement policy without expanding this focused fix into shell evaluation.

## Verification

- `python .github/scripts/test_hook_guards.py`: 142 assertions passed.
- All 15 canonical Python script gates passed.
- `python .github/scripts/check-plugin-refs.py` passed.
- `python -m py_compile plugins/ca/hooks/pre-bash.py` passed.
- Hook unittest suite passed 967 tests from clean-checkout conditions.
- Required coverage audit: initial HIGH bypass fixed; re-review PASS with no BLOCK findings.

Closes #335


SUaDtL:

PR #351, opened 2026-07-20
docs(context): require isolated scout reports

SUaDtL:
## Summary

- make the restricted scout role canonical for brownfield context creation
- require six isolated scout reports and block on hosts that cannot provide isolated subagents
- scope the small-repository inline shortcut and direct source reads to decision-variance
- document the context-creation exception in Codex host notes and the public scout reference
- close the stale task/standup site-mirror task with evidence from merged commit f51063d
- close v2.docs.0004

## Why block instead of inline

Context creation deliberately synthesizes from reports so the orchestrator never loads raw auth, infrastructure, or database source into its own context. An inline fallback cannot provide that isolation. The portable contract is therefore capability-based: dispatch isolated scouts or stop before Phase 2.

## Review

Coverage, architecture, and security reviewers each found a contract edge in the first pass. The final diff resolves the small-repository threshold conflict, removes the impossible Codex inline-isolation claim, and avoids a host-specific fallback in the canonical template. All three re-reviews report no remaining findings.

## Verification

- 16 repository script suites passed
- 967 hook tests passed
- 403 site tests passed
- site TypeScript typecheck and production build passed
- Claude and Codex generated surfaces are synchronized
- canonical hook core is synchronized across both host copies
- plugin reference graph, whitespace, and secret scans passed


SUaDtL:

PR #357, opened 2026-07-20
test: isolate NO_COLOR palette equality from session age

### LINKED ISSUE #355: ci: NO_COLOR palette equality test races live age boundary
## Summary

`TestActivePaletteCompatibility.test_no_color_makes_every_builtin_layout_byte_identical` can fail when its five sequential statusline subprocesses cross a live age-display boundary. The assertion attributes every byte difference to theme selection, but the payload leaves the ledger-derived `Today ... age` field live.

## Evidence

Observed locally on WSL/Ubuntu while verifying #352:

```
FAIL: test_no_color_makes_every_builtin_layout_byte_identical
  self.assertTrue(all(output == outputs[0] for output in outputs[1:]))
```

The exact isolated rerun failed again. Capturing each rendered output then showed identical theme/layout bytes once all five executions landed on the same side of the boundary; output included the dynamic row:

```
Today ... age 1m
```

The test launches five subprocesses sequentially with the same session id and temp workspace, so a transition such as `age 59s` to `age 1m` makes outputs differ even under `NO_COLOR=1`.

## Expected

The test should hold all non-theme inputs constant and prove only that removing color makes the built-in themes byte-identical.

## Suggested regression

Freeze or isolate the ledger/age input for all five subprocesses, then exercise a forced boundary to prove live time cannot affect the comparison.

## Scope

Pre-existing test reliability defect; no product behavior change requested. Discovered while fixing #352 and intentionally excluded from that narrow patch.

SUaDtL:
## Summary

Keep the `NO_COLOR` palette equality test independent of mutable user-ledger state. The theme-only payload now omits its unused session id, so every renderer subprocess produces a call-local `age 0s` instead of racing a live age boundary.

Shipped statusline behavior is unchanged.

Closes #355

## Verification

- Deterministic red proof reproduced `age 59s` versus `age 1m` as the sole byte difference.
- Target test passed 30 of 30 fresh-process repetitions on Linux.
- Full hook suite passed 967 tests with 1 intentional skip on Linux.
- All 25 `.github/scripts/test_*.py` suites passed.
- Python compile, whitespace validation, and staged secrets scan passed.
- Independent coverage review passed with no findings.

## Tradeoff

The comparison no longer exercises session-dependent ledger output. Under the project conflict hierarchy, that behavior is outside this palette test's obligation and remains covered by dedicated ledger and statusline tests; keeping it here made a theme assertion depend on wall-clock timing.

### DIFF
--- a/plugins/ca/hooks/tests/test_colorlib.py
+++ b/plugins/ca/hooks/tests/test_colorlib.py
@@ -396,8 +396,9 @@ if sl.V2 != sl._colorlib.V2 or sl._boxlib._V0 != sl._colorlib.V0 \
 
     def test_no_color_makes_every_builtin_layout_byte_identical(self):
         renderer = os.path.join(_HOOKS_DIR, "statusline.py")
-        payload = json.dumps({"session_id": "palette-layout-test",
-                              "workspace": {"current_dir": self.tmp_path()},
+        # Keep this theme-only comparison independent of the mutable user ledger.
+        # Without a session id, every subprocess renders a call-local 0s age.
+        payload = json.dumps({"workspace": {"current_dir": self.tmp_path()},
                               "model": {"display_name": "Test"}})
         outputs = []
         for theme in ("violet", "blue", "green", "amber", "mono"):


SUaDtL:

Issue #312, opened 2026-07-15
tribunal: tokens_actual can be populated from per-agent subagent transcripts (currently always null)


SUaDtL:
## Summary

The tribunal telemetry payload treats `tokens_actual` as *"optional/best-effort — null when the orchestrator could not observe subagent spend"* (`references/telemetry.md`), and `schemas.md` says a `lens-completed` `tokens` value is recorded *"when the orchestrator can see it; null/omitted when unobserved."* In practice the orchestrator never can see it during the run, so `tokens_actual` is effectively always null — which means the `tokens_estimated` vs `tokens_actual` calibration loop (the whole point of that pair) never gets real data.

This is a limitation, not a crash: the estimator can't self-calibrate because it never receives an actual.

## The recoverable signal

On the Claude Code harness, a dispatched subagent's token spend IS durably recorded — just not where you'd first look:

- It does **not** appear as `isSidechain: true` lines in the main session transcript (that flag isn't set for Task/Agent subagents).
- Each dispatched agent writes its **own** transcript file at `<session-dir>/subagents/agent-<agentId>.jsonl`, one `message.usage` object per line (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`), with a companion `agent-<agentId>.meta.json` recording `agentType`, the dispatch `description`/label, and `toolUseId`.
- `<session-dir>` is the dir beside the running `<session-id>.jsonl`.

## Fix action

After a lens returns, before writing its `lens-completed` event, sum that agent's own transcript:

```sh
jq -s '[.[]|.message.usage//empty]
       | {in:(map(.input_tokens//0)|add),
          out:(map(.output_tokens//0)|add),
          cache_read:(map(.cache_read_input_tokens//0)|add)}' \
  "<session-dir>/subagents/agent-<agentId>.jsonl"
```

Record the dispatched `agentId` on the `lens-launched` event so the lens→file map is explicit (or match by the `description` label in `.meta.json`, since each lens is dispatched once). Attribution is then **exact per lens** — one file per dispatch — not a best-effort time-window guess, and both the run total and the per-lens split are exact.

## Bonus for the estimator

[...]

PR #358, opened 2026-07-20
fix(tribunal): recover Codex reviewer token usage

### LINKED ISSUE #311: Codex tribunal does not emit reviewer token usage into lens-completed events
## Summary

A Codex-hosted `$ca-tribunal` run completed every reviewer lens without emitting the optional `tokens` field in any `lens-completed` event. The generated telemetry therefore records `tokens_actual: null`, so the estimator cannot self-calibrate for Codex runs.

## Observed

- codeArbiter Codex plugin: `0.3.0`
- Host and orchestrator model: Codex / `gpt-5-codex`
- Completed lenses: 11
- `lens-completed` events containing `tokens`: 0 of 11
- Estimated tokens: `3015018`
- Actual tokens: `null`
- Public scrubbed reproduction: #310

The tribunal run exposed each lens model, `surface_seen`, and finding count, but no token-usage value. No repository identity, paths, code, or finding text are needed to reproduce the reporting gap.

## Expected

When Codex exposes reviewer usage, the host integration should record it on each `lens-completed` event so `tokens_actual` can sum the observed values. If the Codex agent surface does not expose usage, the run should record an explicit unavailability reason or capability state instead of leaving the cause implicit in a null value.

## Impact

Codex tribunal estimates cannot be calibrated against measured consumption. Maintainers also cannot distinguish an unsupported host surface from a transient instrumentation failure by inspecting telemetry alone.

## Acceptance criteria

- Codex tribunal runs record numeric reviewer token usage when the host exposes it.
- Telemetry sums the recorded values into `tokens_actual`.
- When usage is unavailable, run state distinguishes an unsupported host capability from missing or failed instrumentation.
- Automated coverage exercises the Codex unavailable-usage path and the numeric aggregation path.

SUaDtL:
## Summary

- Recover cumulative reviewer usage from the exact Codex agent session.
- Record complete, partial, or unavailable actuals with a closed reason taxonomy.
- Bound and confine transcript parsing so malformed local state fails soft and cannot leak arbitrary text into telemetry.
- Correct the Codex host note to reflect current host-provided subagents.

## Why

Codex dispatch results do not include usage, which left tribunal actuals blank. The best exact-thread source is local session telemetry, but its JSONL format is not stable. This isolates parsing behind a fail-soft helper and makes format drift explicit instead of silently reporting no actuals.

## Verification

- `python -m unittest plugins.ca.hooks.tests.test_codex_tribunal_usage -v` (22 passed)
- Full hook suite under WSL (989 passed, 1 expected skip)
- All 25 `.github/scripts/test_*.py` suites
- `python tools/build-surface.py --check`
- `python .github/scripts/check-plugin-refs.py`
- Live exact-thread receipt recovery
- Independent coverage and security reviews: PASS

## Tradeoff

The local session format is an unstable host detail. The helper treats it as best-effort, reads only metadata and cumulative usage events, applies strict bounds and path confinement, and returns fixed unavailable states instead of blocking the tribunal.

Closes #311

Issue #312 remains the separate follow-up for broader transcript recovery on the other host.


SUaDtL:

PR #359, opened 2026-07-20
fix(tribunal): recover Claude reviewer token usage

### LINKED ISSUE #312: tribunal: tokens_actual can be populated from per-agent subagent transcripts (currently always null)

SUaDtL:
## Summary

- Recover exact per-lens Claude token usage from the documented subagent transcript location.
- Report fresh input, cache creation, cache read, output, and total tokens with explicit unavailable reasons on changed or unsafe host state.
- Keep parsing bounded, path-confined, fail-soft, and privacy-safe.
- Validate returned agent IDs before any shell command is constructed.

Claude documents the per-agent transcript location in its [subagent documentation](https://code.claude.com/docs/en/sub-agents).

## Stacked dependency

This PR is based on `fix/tribunal-token-telemetry` from #358 so the Claude and Codex host implementations share the same schema and release surface. It should be retargeted to `main` after #358 lands.

## Verification

- 34 focused Claude and Codex usage tests pass.
- Full hook test suite passes with the existing intentional skip.
- All 25 repository script suites pass.
- Generated surfaces and shared core are synchronized.
- Claude and Codex reference graphs, badge/catalog, license, JSON manifest, syntax, whitespace, and secret checks pass.
- Live exact-agent recovery returns 3,947,997 tokens with all four components.
- Independent coverage review passes with no findings.
- Independent security and privacy re-review passes with zero findings.

Closes #312


SUaDtL:

PR #420, opened 2026-07-24
fix(security): isolate Pi child credentials and make farm writes symlink-safe

### LINKED ISSUE #372: [critical] Isolate Pi child HOME from the operator credential store
# Isolate Pi child HOME from the operator credential store

**Severity:** critical | **Confidence:** 0.97 | **Effort:** L

**Where:**
- plugins/ca-pi/tools/src/child-env.ts:9-21
- plugins/ca-pi/tools/src/child-env.ts:83-102
- plugins/ca-pi/tools/src/tool-guard.ts:115-120
- plugins/ca-pi/tools/src/child-extension.ts:197-218

**Evidence:** `buildChildEnv()` copies the parent `HOME`, Windows profile directories, and `PI_CODING_AGENT_DIR` into every child. The child installs normal bash/read tools, and `INSPECTION_COMMANDS` automatically allows `cat`/`Get-Content`. Executing the current classifiers produced `{"actions":["inspection"],"verdict":{"decision":"allow"}}` for `Get-Content C:\\Users\\operator\\.pi\\agent\\auth.json`; executing `buildChildEnv` preserved `/home/operator/.pi/agent`.

**Impact:** A delegated child model can read the operator's real Pi auth store and other profile files without a permission prompt, exposing credentials for providers unrelated to the selected child provider and defeating the documented provider-specific minimal-environment boundary.

**Recommendation:** Launch every child with a fresh private HOME and Pi agent/session directories outside the operator profile. Admit only the selected provider's scoped credential mechanism; if Pi cannot provide a child-scoped opaque credential capability for auth-store-only providers, fail those child launches closed rather than sharing the operator store.

**Acceptance criteria:**
- Child `HOME`, profile directories, `PI_CODING_AGENT_DIR`, and `PI_CODING_AGENT_SESSION_DIR` resolve to fresh child-owned locations that contain no operator auth store.
- A live child cannot read a sentinel placed in the operator Pi auth directory through bash or read tools.
- Only the selected provider's reviewed credential path is available to the child; unrelated provider credentials remain absent.
- Temporary child credential/state directories are removed on success, failure, cancellation, timeout, and parent shutdown without inspecting or mutating the operator store.

### LINKED ISSUE #373: [critical] Make farm worker writes symlink-safe at the filesystem boundary
# Make farm worker writes symlink-safe at the filesystem boundary

**Severity:** critical | **Confidence:** 0.95 | **Effort:** M

**Where:**
- plugins/ca/tools/farm.ts:752-783
- plugins/ca/tools/farm.ts:886-905
- plugins/ca/tools/farm.ts:1054-1066

**Evidence:** `runWorker()` computes `absPath = path.resolve(cwd, cleanPath)`, checks only lexical `isInside(cwd, absPath)`, then calls `mkdir()`/`writeFile()` without checking any existing path component for symlinks. `postApplySweep()` repeats the same lexical check, and best-of-N materialization uses the same pattern. A tracked symlink or junction below `cwd` is therefore inside lexically while its write target is outside physically.

**Impact:** Untrusted farm model output can overwrite an arbitrary same-user host file reachable through a pre-existing worktree symlink/junction. Because writing through the link does not modify the link itself, the subsequent Git drift sweep can miss the external mutation entirely.

**Recommendation:** Replace path-string containment with a symlink-safe writer: canonicalize and verify every existing ancestor beneath the canonical worktree, reject symlink/reparse-point targets and parents, and create/open the destination with no-follow semantics where supported. Apply the same primitive to single-worker, best-of-N, and mutation restore writes.

**Acceptance criteria:**
- A regression test places a file symlink and directory symlink/junction inside a worktree pointing to an external sentinel and proves every farm write path rejects them without changing the sentinel.
- All existing ancestors and the destination are verified against the canonical worktree immediately before open/write, including Windows reparse-point behavior.
- Single-worker output, best-of-N winner materialization, and mutation write/restore paths share the same symlink-safe filesystem primitive.
- Containment failure escalates the task with a bounded diagnostic and stages no files.

SUaDtL:
Lands the deferred security commit `980688f`, which was authored on the local-only branch `codex/tribunal-critical-fixes` and **never pushed**. Its Pi sibling work reached main via #313 (`feat/pi-support`), but this security commit did not — verified: `child-env.ts` on main is 122 lines with zero credential projection, and ADR-0016 was absent from `.codearbiter/decisions/`.

Closes #372
Closes #373

## What lands

**ADR-0016 — bounded selected-provider credential projection** (supersedes the opaque-auth clause of ADR-0014). Corroborated by `DECISION-0021` in the decision log, dated 2026-07-22, `Decided by: SUaDtL@... ("1")` — the maintainer's literal conflict-option selection, matching the established `DECISION-0016 ("a")` format. Re-confirmed by the maintainer before this rebase.

- **Pi child credentials** — `ca-pi` reads the operator `auth.json` only while preparing a child, under a strict size bound, and projects **only the one selected provider record** into a fresh private ephemeral root. Never in argv, prompts, results, logs, telemetry, or `.codearbiter/`. Scrubbed on every path including failure; unprovable cleanup returns a bounded degraded failure.
- **Farm worker writes** — new `worktree-fs.ts` primitive makes every model-controlled write reject symlink, junction, hardlink, and replacement races before mutation.

## Verification

Cherry-picked onto current main with **no conflicts**. Both committed bundles (`plugins/ca/tools/farm.js`, `plugins/ca-pi/extensions/codearbiter.js`) were rebuilt from source and are **byte-identical** to what the commit carried, so neither is stale against post-#313 main.

| Suite | Before | After |
|---|---|---|
| ca-pi (vitest) | 545 passed | **551 passed**, 1 skipped |
| ca-farm (vitest) | 198 passed | **207 passed**, 1 skipped |
| python-hooks | 1057 OK | **1057 OK** |

`tsc --noEmit` clean in both TS packages. `git diff --check` clean (LF preserved).

## Conflict hierarchy

Resolved at **Level 1** (security & audit-trail correctness) — the credential boundary and the audit-attributed ADR both sit above velocity concerns.

## Also included

Two audit-trail records left uncommitted by prior sessions: a `/ca:dev` enter/exit pair in `overrides.log`, and an `H-01` branch-guard block in `gate-events.log` (the gate correctly refusing a commit evaluated on main during this session).


SUaDtL:
Superseded and split.

Landing `980688f` on current main surfaced two things:

1. **A contract conflict (resolved).** `test_pi_child_live.py` asserted that none of the five Pi child sources may reference `auth.json` — the ADR-0014 opaque-auth clause. ADR-0016 supersedes exactly that clause and sanctions `child-env.ts` as a narrow secret-transport boundary, so the blanket ban contradicted the accepted decision and failed all six adapter-contract matrix cells. Fix drafted here: narrow the ban to the four non-transport sources and replace it, for `child-env.ts`, with positive assertions of ADR-0016's bounded contract (size bound, private `mkdtemp` root, exact selected-provider record, scrub set, credential truncation). Every assertion was mutation-tested and bites.

2. **A live regression (unresolved).** With that cleared, all six cells still fail — now at the *live* child contract:
   ```
   live child did not complete through production runner:
   result={'terminal': 'degraded', 'diagnostic': 'Pi child isolation failed safely; no inline promotion is available; run /ca-doctor.'}
   provider_turns=0 runner_stderr=''
   ```
   The credential-projection path fails closed against a real Pi 0.80.5/0.80.10 runtime with no reason string. This is a genuine defect in the projection code as authored on the stale branch — and the likely reason the commit was never pushed.

The farm half of `980688f` (#373) is independent and green, so it ships separately as #425.

The Pi half (#372, ADR-0016) is held until the live-child regression is fixed. Reopening as a fresh PR once it passes the live contract.

PR #431, opened 2026-07-25
fix(hooks): stop losing and corrupting durable state in three core readers

### LINKED ISSUE #396: Preserve a retryable dev-exit record when audit append fails
# Preserve a retryable dev-exit record when audit append fails

**Severity:** high | **Confidence:** 0.99 | **Effort:** M

**Where:**
- core/pysrc/session-start.py:527-615

**Evidence:** `clear_dev_marker` catches and discards an `OSError` from appending the synthetic `DEV: exit` at lines 606-611, then independently removes `dev-active` at lines 612-615. A bounded monkeypatched reproduction forced only the overrides append to raise `OSError('locked')`; the function returned normally, deleted the marker, and left `overrides.log` containing only `DEV: enter`.

**Impact:** A transient lock, disk error, or permission failure permanently erases the only retry signal after an abandoned maintainer session. The append-only audit trail is left with an unmatched `DEV: enter`, contradicting the documented guarantee that SessionStart self-heals the trail, and no later session can know that the close row is still owed.

**Recommendation:** Keep startup fail-open but make close logging retryable and ordered. Persist a minimal atomic pending-close record (or retain/transition the marker) before attempting the append; remove that durable retry state only after the append is confirmed. Deduplicate retries with a stable close identity so a crash after append but before marker cleanup cannot create ambiguous duplicate exits.

**Acceptance criteria:**
- When the overrides append fails, SessionStart still returns successfully but leaves a durable retryable close record.
- A later SessionStart appends the missing `DEV: exit` exactly once and then removes the retry state.
- A crash/restart at every boundary between pending-record creation, audit append, and cleanup yields one attributable close row without losing the enter/exit pairing.
- Tests inject append failure, marker cleanup failure, and post-append crash behavior for the canonical source and generated host copies.

### LINKED ISSUE #410: Validate provenance JSON records before admitting them to the store
# Validate provenance JSON records before admitting them to the store

**Where:**
- core/pysrc/_provenancelib.py:228-239
- core/pysrc/_provenancelib.py:603-628

**Evidence:** `read_provenance()` documents `dict | None` but returns the result of `json.load()` without a shape check. `load_provenance_dir()` then calls `record.get()` inside one catch around the entire directory loop. In a live temporary directory containing `a.json = []` followed by a valid `b.json`, `read_provenance(a)` returned `[]` and `load_provenance_dir()` returned `{}`, silently dropping the valid record as well.

**Impact:** One syntactically valid but schema-invalid provenance file can abort the remaining directory scan and suppress otherwise valid source evidence. SessionStart drift reporting and commit-gate auto-heal can therefore operate from an incomplete provenance map without identifying the corrupt record.

**Recommendation:** Make `read_provenance()` validate the complete provenance record shape and return None for every non-mapping or invalid version. In the directory loader, isolate each file in its own error boundary so one bad record cannot suppress later records, and surface a bounded warning where the caller supports it.

**Acceptance criteria:**
- `read_provenance()` returns None for null, arrays, scalars, mappings with the wrong schema, and malformed entries.
- A directory containing invalid and valid records loads every valid record regardless of filename order.
- The loader never raises on a malformed record and records or returns a bounded corruption diagnostic without exposing file contents.
- Tests cover mixed directories with every JSON top-level type and a valid record after each invalid record.

### LINKED ISSUE #413: Make the subagent JSONL reader honor one stable result contract
# Make the subagent JSONL reader honor one stable result contract

**Severity:** low | **Confidence:** 1 | **Effort:** S

**Where:**
- core/pysrc/_subagentslib.py:91-109
- core/pysrc/_subagentslib.py:120-162
- core/pysrc/statusline.py:647-654

**Evidence:** `read_subagents()` documents and normally returns a four-item tuple, but its directory-error branch returns three items. It also decodes each JSONL line and immediately calls `.get()` without requiring an object. Live temporary-directory calls reproduced `(0, [], (0, 0))` for a missing directory and `AttributeError: 'list' object has no attribute 'get'` for the valid JSON line `[]`. The statusline then either fails tuple unpacking or has the exception swallowed by `safe()`, hiding the entire subagent section.

**Impact:** A normal directory race or one syntactically valid non-object event can erase all subagent visibility from the rich statusline without a useful diagnostic. Direct callers also cannot rely on the function's documented tuple shape, making future reuse and extension modes prone to unpacking failures.

**Recommendation:** Define one explicit result shape for `read_subagents()` and return it on every path. Parse JSONL values as unknown, skip non-mapping records before field access, and keep file-local corruption isolated so later valid records still contribute.

**Acceptance criteria:**
- Every return path produces the documented four-item result with stable element types.
- Missing directories, directory races, and per-file stat/open failures return a valid empty or partial result without exception.
- Null, array, scalar, malformed, and object JSONL records can coexist; invalid records are skipped and later valid records are still counted.
- Statusline tests cover the missing-directory path and mixed-shape JSONL without suppressing valid subagent rows.

SUaDtL:
Three durability defects in the canonical `core/pysrc/` hook layer. Each is the
same shape: the *degradation* path destroys or drops state instead of preserving
it. All three edits are in `core/pysrc/`, re-vendored into `plugins/ca`,
`plugins/ca-codex` and `plugins/ca-pi` via `python tools/sync-core.py`.

Closes #396
Closes #413
Closes #410

---

## #396 — a retryable dev-exit record when the audit append fails

`clear_dev_marker` appended the synthetic `DEV: exit` under a fire-and-forget
`except OSError: pass`, then removed `dev-active` from a **separate** try block.
A locked file, full disk, or permission blip therefore permanently erased the
only retry signal and left the append-only trail with an unmatched `DEV: enter`
— contradicting the documented SessionStart self-heal guarantee.

The close now runs through one settlement step (`_settle_dev_close`) built
around a small write-ahead record at `.codearbiter/.markers/dev-close-pending.json`:

```json
{"lines": ["<owed DEV: exit line>"], "marker_mtime": 1753405.2}
```

- `lines` — close lines still owed to `overrides.log`, emptied one at a time as
  each append is **confirmed**. Bounded to 8.
- `marker_mtime` — the identity of the `dev-active` marker the close belongs to.
  While the record still names a **live** marker it doubles as a tombstone, so a
  failed `os.remove` cannot make the next session mint a *second* close row.

Every crash boundary is covered:

| crash point | state on disk | next SessionStart |
|---|---|---|
| before the append | record present, line owed | replays the append |
| after the append, before cleanup | record present, line owed | bounded 64 KiB tail scan sees the line already landed, drops it instead of duplicating |
| marker removal fails | record present, no line owed | only retries the removal |

Startup still fails **OPEN** — `clear_dev_marker` returns normally on every
path, exactly as before. The two early returns that leave a *live* marker owned
by another session deliberately do **not** settle: discarding a tombstone for a
marker you are not entitled to touch is precisely how a duplicate close gets
minted.

## #413 — one stable result contract for the subagent JSONL reader

Two independent defects, both of which erase the entire subagent section of the
rich statusline with no diagnostic:

1. `read_subagents` returned a **3-item** tuple on its directory-error branch
   while the module docstring, the happy path, and `statusline.py`'s own
   import-failure stub all promise **4**. `statusline.py` unpacks four names
   *outside* `safe()`, so a routine directory race was a hard `ValueError`.
2. `json.loads(ln)` was followed immediately by `d.get("message")` inside a try
   that caught `OSError` only — so one syntactically valid non-object line
   (`[]`, `null`, `3`, `"s"`) raised `AttributeError` out of the reader.

Fixed: one `empty = (0, 0, [], (0, 0))` returned on every error path, an
`isinstance(d, dict)` skip before any field access, and the per-file boundary
widened from `OSError` to `Exception` so file-local corruption costs only its
own row — matching the module's own stated contract ("Never raise on malformed
input — every reader degrades to a safe blank").

## #410 — validate provenance records before admitting them

`read_provenance` documented `dict | None` but returned the raw `json.load`
result, so a file containing `[]` was admitted to the store *as a list*.
`load_provenance_dir` then wrapped the **entire** glob loop in one try/except,
so the first `record.get("doc")` `AttributeError` aborted the scan and silently
dropped every later valid record — SessionStart drift reporting and commit-gate
auto-heal then ran against a truncated provenance map with no indication
anything was missing.

- New `valid_provenance_record()` checks the **top-level** v1 shape only
  (`schema == 1` with `bool` excluded since `True == 1`, plus `doc: str`,
  `created: str`, `interview_derived: bool`, `entries: list`). Entry-level
  tolerance stays where it already lives (`compute_drift`, `provenance_pointer`),
  so one odd claim never costs a doc its whole record.
- `read_provenance` returns `None` for anything that fails it.
- `load_provenance_dir` gets a **per-file** error boundary plus an optional
  `skipped=[]` list collecting up to `MAX_SKIPPED_REPORTED` (5) rejected
  **basenames** — names only, never file contents.

Verified against the real writers: every provenance record in the repo and in
the surface prose is produced by `new_record` / `write_provenance` /
`write_stub` / `rebaseline`, all of which emit the canonical shape. No
hand-authored provenance JSON exists, so the validation admits everything that
was already valid.

---

### #396 — `plugins/ca/hooks/tests/test_session_start.py::TestDevExitRetryablePendingClose`

```
F.FFFF
======================================================================
FAIL: test_append_failure_leaves_a_durable_retryable_close_record (test_session_start.TestDevExitRetryablePendingClose.test_append_failure_leaves_a_durable_retryable_close_record)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_session_start.py", line 627, in test_append_failure_leaves_a_durable_retryable_close_record
    self.assertTrue(os.path.isfile(self.pending),
                    "a failed append must leave a durable pending-close record")
AssertionError: False is not true : a failed append must leave a durable pending-close record

======================================================================
FAIL: test_corrupt_pending_record_is_discarded_not_jammed (test_session_start.TestDevExitRetryablePendingClose.test_corrupt_pending_record_is_discarded_not_jammed)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_session_start.py", line 703, in test_corrupt_pending_record_is_discarded_not_jammed
    self.assertFalse(os.path.isfile(self.pending))
AssertionError: True is not false

======================================================================
FAIL: test_crash_after_append_before_cleanup_does_not_duplicate (test_session_start.TestDevExitRetryablePendingClose.test_crash_after_append_before_cleanup_does_not_duplicate)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_session_start.py", line 664, in test_crash_after_append_before_cleanup_does_not_duplicate
    self.assertFalse(os.path.isfile(self.pending),
                     "the settled record must be cleared")
AssertionError: True is not false : the settled record must be cleared

======================================================================
FAIL: test_later_session_appends_the_missing_exit_exactly_once (test_session_start.TestDevExitRetryablePendingClose.test_later_session_appends_the_missing_exit_exactly_once)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_session_start.py", line 642, in test_later_session_appends_the_missing_exit_exactly_once
    self.assertEqual(len(self._exit_lines()), 1,
                     "the owed DEV: exit must land exactly once on retry")
AssertionError: 0 != 1 : the owed DEV: exit must land exactly once on retry

======================================================================
FAIL: test_marker_removal_failure_does_not_duplicate_the_close_row (test_session_start.TestDevExitRetryablePendingClose.test_marker_removal_failure_does_not_duplicate_the_close_row)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_session_start.py", line 684, in test_marker_removal_failure_does_not_duplicate_the_close_row
    self.assertEqual(len(self._exit_lines()), 1,
                     "the same marker must not be closed twice in the audit trail")
AssertionError: 2 != 1 : the same marker must not be closed twice in the audit trail

----------------------------------------------------------------------
Ran 6 tests in 0.027s

FAILED (failures=5)
```

`0 != 1` is the defect verbatim: the owed close is gone forever. `2 != 1` is the
duplicate-close mode. The sixth test (`test_clean_close_leaves_no_pending_record_behind`)
passes red — it is the guard proving the happy path was never broken.

### #413 — `plugins/ca/hooks/tests/test_statusline.py::TestSubagentResultContract`

```
======================================================================
ERROR: test_missing_directory_unpacks_the_same_way_as_a_real_read (test_statusline.TestSubagentResultContract.test_missing_directory_unpacks_the_same_way_as_a_real_read)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_statusline.py", line 924, in test_missing_directory_unpacks_the_same_way_as_a_real_read
    active, recent, shown, (tin, tout) = subs.read_subagents(gone)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 4, got 3)

======================================================================
ERROR: test_non_object_jsonl_records_are_skipped_not_raised (test_statusline.TestSubagentResultContract.test_non_object_jsonl_records_are_skipped_not_raised)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_statusline.py", line 952, in test_non_object_jsonl_records_are_skipped_not_raised
    active, recent, shown, (tin, tout) = subs.read_subagents(td)
  File "...\plugins\ca\hooks\_subagentslib.py", line 132, in read_subagents
    msg = d.get("message")
          ^^^^^
AttributeError: 'list' object has no attribute 'get'

======================================================================
ERROR: test_statusline_renders_through_a_subagent_directory_race (test_statusline.TestSubagentResultContract.test_statusline_renders_through_a_subagent_directory_race)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_statusline.py", line 931, in test_statusline_renders_through_a_subagent_directory_race
    out = sl.render(json.dumps({"session_id": "sid"}))
  File "...\plugins\ca\hooks\statusline.py", line 535, in render
    return _render_active_palette(raw)
  File "...\plugins\ca\hooks\statusline.py", line 653, in _render_active_palette
    active, recent, shown, (tin, tout) = res
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 4, got 3)

======================================================================
FAIL: test_missing_directory_returns_the_documented_four_item_result (test_statusline.TestSubagentResultContract.test_missing_directory_returns_the_documented_four_item_result)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\plugins\ca\hooks\tests\test_statusline.py", line 911, in test_missing_directory_returns_the_documented_four_item_result
    self.assertEqual(len(res), 4,
                     "every return path must produce the documented 4-item result")
AssertionError: 3 != 4 : every return path must produce the documented 4-item result

----------------------------------------------------------------------
Ran 5 tests in 0.061s

FAILED (failures=1, errors=4)
```

The third trace is the real-world consequence reproduced end to end: a
`ValueError` thrown from `statusline.py` line 653, exactly the line the issue
names.

### #410 — `.github/scripts/test_provenancelib.py::ProvenanceRecordShapeTest`

```
FFFFFEF.EFF.
======================================================================
ERROR: test_corruption_diagnostic_is_bounded (test_provenancelib.ProvenanceRecordShapeTest.test_corruption_diagnostic_is_bounded)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\.github\scripts\test_provenancelib.py", line 2716, in test_corruption_diagnostic_is_bounded
    pl.load_provenance_dir(d, skipped=skipped)
TypeError: load_provenance_dir() got an unexpected keyword argument 'skipped'

======================================================================
FAIL: test_a_valid_record_survives_after_every_invalid_top_level_type (test_provenancelib.ProvenanceRecordShapeTest.test_a_valid_record_survives_after_every_invalid_top_level_type) (raw='[]')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\.github\scripts\test_provenancelib.py", line 2688, in test_a_valid_record_survives_after_every_invalid_top_level_type
    self.assertIn("tech-stack", loaded,
                  f"a valid record after {raw!r} must still load")
AssertionError: 'tech-stack' not found in {} : a valid record after '[]' must still load

======================================================================
FAIL: test_invalid_record_does_not_drop_the_valid_records_after_it (test_provenancelib.ProvenanceRecordShapeTest.test_invalid_record_does_not_drop_the_valid_records_after_it)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\.github\scripts\test_provenancelib.py", line 2678, in test_invalid_record_does_not_drop_the_valid_records_after_it
    self.assertEqual(sorted(loaded), ["code-map", "tech-stack"],
                     "one invalid record must not abort the directory scan")
AssertionError: Lists differ: [] != ['code-map', 'tech-stack']

Second list contains 2 additional elements.
First extra element 0:
'code-map'

- []
+ ['code-map', 'tech-stack'] : one invalid record must not abort the directory scan

======================================================================
FAIL: test_read_provenance_rejects_a_mapping_with_the_wrong_schema (test_provenancelib.ProvenanceRecordShapeTest.test_read_provenance_rejects_a_mapping_with_the_wrong_schema)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\.github\scripts\test_provenancelib.py", line 2661, in test_read_provenance_rejects_a_mapping_with_the_wrong_schema
    self.assertIsNone(
        pl.read_provenance(path),
        f"read_provenance must return None for the {name!r} record",
    )
AssertionError: {} is not None : read_provenance must return None for the 'empty' record

======================================================================
FAIL: test_read_provenance_rejects_every_non_mapping_top_level_type (test_provenancelib.ProvenanceRecordShapeTest.test_read_provenance_rejects_every_non_mapping_top_level_type)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "...\.github\scripts\test_provenancelib.py", line 2637, in test_read_provenance_rejects_every_non_mapping_top_level_type
    self.assertIsNone(
        pl.read_provenance(path),
        f"read_provenance must return None for a top-level {raw!r}",
    )
AssertionError: [] is not None : read_provenance must return None for a top-level '[]'

----------------------------------------------------------------------
Ran 8 tests in 0.047s

FAILED (failures=8, errors=2)
```

`'tech-stack' not found in {}` is the silent-drop defect: the loader returned an
empty map even though a perfectly valid record sat next to the bad one.

---

## Suite results

| suite | before (origin/main @ 4ec1a8a) | after |
|---|---|---|
| `python -m unittest discover -s plugins/ca/hooks/tests -p "test_*.py"` | 1057 tests, **OK** | 1068 tests, **OK** (+11) |
| `cd .github/scripts && python -m unittest discover -s . -p "test_*.py"` | 1064 tests, 2 failures / 2 errors / 5 skipped | 1072 tests, **same** 2 failures / 2 errors / 5 skipped (+8) |
| `python tools/sync-core.py --check` | OK | OK — 47 core files x 3 plugins, byte-identical |
| `python tools/build-host-packages.py --check --release-guard-base <merge-base>` | n/a | OK — `0.1.2 -> 0.1.3` advanced together |
| `check_badge_consistency.py` / `check_docs_contract.py` / `check-plugin-refs.py` | — | all clean |
| `git diff --check` | — | clean; all touched files LF |

The four `.github/scripts` failures are pre-existing on `origin/main` and
unrelated — see NEEDS-TRIAGE below.

No TypeScript package was touched, so `plugins/ca/tools` (farm) and
`plugins/ca-sandbox/tools` were not rebuilt — `farm.js` / `sandbox.js` are
unchanged and the staleness gate is unaffected.

### Payload version gate

`plugins/ca-pi/hooks/**` changed (vendored core sync), and the Pi guard is
stricter than ca/ca-codex — it requires the version to advance regardless of
whether a tag is published. So **ca-pi 0.1.2 → 0.1.3**, with the regenerated
root `package.json` (`python tools/build-host-packages.py`) and a new
`## [0.1.3] - 2026-07-24` heading, all in one `chore(ca-pi)` commit.

`ca` (2.9.1) and `ca-codex` (0.3.0) need no bump — neither `v2.9.1` nor
`ca-codex-v0.3.0` is published, and both guards only bite once the tag exists.
Confirmed against `git ls-remote --tags origin`.

---

## NEEDS-TRIAGE

**1. `test_pi_benchmark` is red on `origin/main` before this branch exists
(environmental, not a regression).** Four tests fail identically at the
`4ec1a8a` baseline and after this change:

```
ERROR: test_measurements_use_exact_public_shape_and_100_warm_events
ERROR: test_pi_record_does_not_load_the_generated_python_host_adapter
FAIL:  test_cli_emits_only_three_bounded_json_records
FAIL:  test_cli_real_spawn_flag_adds_a_fourth_labeled_record
```

Root cause is a missing local toolchain, not the code:

```
File ".github/scripts/pi_benchmark.py", line 151, in _production_pi_samples
    raise RuntimeError("Pi benchmark requires the installed pinned esbuild")
```

CI presumably installs the pinned esbuild, so this is likely green there. Worth
a separate look at whether these should skip rather than error when the
toolchain is absent — every local contributor otherwise sees a red
`.github/scripts` suite. Out of scope for this lane; filed here rather than
fixed.

**2. The `dev-close-pending` record is settled only by SessionStart.**
`/ca:arbiter` (the normal dev exit) removes `dev-active` through its own prose
path and does not consult the pending record. That is harmless — the record is
retired by the next SessionStart's no-marker settle — but it means a genuinely
owed close can sit on disk until the next session starts rather than being
flushed at exit time. Deliberate: wiring this into the arbiter prose is a
surface change well outside a `core/pysrc` durability fix.

**3. `load_provenance_dir(skipped=...)` has no caller yet.** The parameter is
opt-in and nothing passes it today — the acceptance criterion asked for a
bounded diagnostic "where the caller supports it", and adding it to
`startup_drift_line` or the context-check surface would change the SessionStart
output line. The plumbing is in place and tested; surfacing it is a follow-up.

### DIFF
--- a/.github/scripts/test_provenancelib.py
+++ b/.github/scripts/test_provenancelib.py
@@ -2597,5 +2597,129 @@ class TimeoutHardeningTest(unittest.TestCase):
         )
 
 
+class ProvenanceRecordShapeTest(unittest.TestCase):
+    """#410: read_provenance documents `dict | None` but used to hand back
+    whatever json.load produced, so a syntactically valid `[]` was admitted to
+    the store as a list. load_provenance_dir then called record.get() inside a
+    single try/except wrapped around the WHOLE glob loop, so the resulting
+    AttributeError aborted the scan and silently dropped every later valid
+    record — SessionStart drift and commit-gate auto-heal then ran against a
+    truncated provenance map with no indication anything was missing."""
+
+    # Every JSON top-level type that is not a provenance record.
+    NON_RECORDS = ("[]", "null", "3", '"a string"', "true", "[{}]")
+
+    def _dir(self):
+        import shutil
+
+        tmp = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, tmp, True)
+        return tmp
+
+    def _write_raw(self, d, name, text):
+        path = os.path.join(d, name)
+        with open(path, "w", encoding="utf-8", newline="\n") as f:
+            f.write(text)
+        return path
+
+    def _valid(self, d, name, doc):
+        record = pl.new_record(doc, created="2026-06-26", entries=[
+            {"path": "package.json", "hash": "a" * 40,
+             "drift_trigger": True, "claims": []},
+        ])
+        pl.write_provenance(os.path.join(d, name), record)
+        return record
+
+    def test_read_provenance_rejects_every_non_mapping_top_level_type(self):
+        d = self._dir()
+        for i, raw in enumerate(self.NON_RECORDS):
+            path = self._write_raw(d, f"nonmap{i}.json", raw)
+            self.assertIsNone(
+                pl.read_provenance(path),
+                f"read_provenance must return None for a top-level {raw!r}",
+            )
+
+    def test_read_provenance_rejects_a_mapping_with_the_wrong_schema(self):
+        d = self._dir()
+        bad = {
+            "empty": {},
+            "no_schema": {"doc": "x", "created": "2026-01-01",
+                          "interview_derived": False, "entries": []},
+            "wrong_version": {"schema": 99, "doc": "x", "created": "2026-01-01",
+                              "interview_derived": False, "entries": []},
+            "schema_as_string": {"schema": "1", "doc": "x", "created": "2026-01-01",
+                                 "interview_derived": False, "entries": []},
+            "entries_not_a_list": {"schema": 1, "doc": "x", "created": "2026-01-01",
+                                   "interview_derived": False, "entries": {}},
+            "doc_not_a_string": {"schema": 1, "doc": 7, "created": "2026-01-01",
+                                 "interview_derived": False, "entries": []},
+            "missing_entries": {"schema": 1, "doc": "x", "created": "2026-01-01",
+                                "interview_derived": False},
+        }
+        for name, record in bad.items():
+            path = self._write_raw(d, f"{name}.json", json.dumps(record))
+            self.assertIsNone(
+                pl.read_provenance(path),
+                f"read_provenance must return None for the {name!r} record",
+            )
+
+    def test_read_provenance_still_accepts_a_canonical_record(self):
+        d = self._dir()
+        record = self._valid(d, "tech-stack.json", "tech-stack")
+        self.assertEqual(pl.read_provenance(os.path.join(d, "tech-stack.json")), record)
+
+    def test_invalid_record_does_not_drop_the_valid_records_after_it(self):
+        # Filenames chosen so the invalid file is globbed FIRST.
+        d = self._dir()
+        self._write_raw(d, "a-bad.json", "[]")
+        self._valid(d, "b-good.json", "tech-stack")
+        self._valid(d, "c-good.json", "code-map")
+        loaded = pl.load_provenance_dir(d)
+        self.assertEqual(sorted(loaded), ["code-map", "tech-stack"],
+                         "one invalid record must not abort the directory scan")
+
+    def test_a_valid_record_survives_after_every_invalid_top_level_type(self):
+        for i, raw in enumerate(self.NON_RECORDS):
+            with self.subTest(raw=raw):
+                d = self._dir()
+                self._write_raw(d, "a-bad.json", raw)
+                self._valid(d, "b-good.json", "tech-stack")
+                loaded = pl.load_provenance_dir(d)
+                self.assertIn("tech-stack", loaded,
+                              f"a valid record after {raw!r} must still load")
+                self.assertEqual(len(loaded), 1)
+
+    def test_loader_never_raises_on_an_all_invalid_directory(self):
+        d = self._dir()
+        for i, raw in enumerate(self.NON_RECORDS):
+            self._write_raw(d, f"bad{i}.json", raw)
+        self.assertEqual(pl.load_provenance_dir(d), {})
+
+    def test_loader_reports_bounded_corruption_without_exposing_contents(self):
+        d = self._dir()
+        # Stand-in for the sort of thing a provenance file really holds —
+        # repo paths and claim prose that has no business in a log line.
+        sensitive = "internal/billing/keyvault.ts"
+        self._write_raw(d, "leaky.json", json.dumps([sensitive]))
+        self._valid(d, "good.json", "tech-stack")
+        skipped = []
+        loaded = pl.load_provenance_dir(d, skipped=skipped)
+        self.assertIn("tech-stack", loaded)
+        self.assertEqual(skipped, ["leaky.json"],
+                         "the loader must name the rejected file, and only its name")
+        self.assertNotIn(sensitive, " ".join(skipped),
+                         "a corruption diagnostic must never carry file contents")
+
+    def test_corruption_diagnostic_is_bounded(self):
+        d = self._dir()
+        for i in range(40):
+            self._write_raw(d, f"bad{i:02d}.json", "[]")
+        skipped = []
+        pl.load_provenance_dir(d, skipped=skipped)
+        self.assertTrue(skipped, "rejected records must be reported")
+        self.assertLessEqual(len(skipped), pl.MAX_SKIPPED_REPORTED,
+                             "the diagnostic must be bounded, not one line per file")
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/core/pysrc/_provenancelib.py
+++ b/core/pysrc/_provenancelib.py
@@ -49,7 +49,11 @@
 #   new_record(doc, *, interview_derived=False, entries=None, created=None)
 #                                        -> dict   canonical record with schema=1
 #   write_provenance(path, record)       -> None   pretty JSON; creates parent dirs
+#   valid_provenance_record(record)      -> bool   True iff a well-formed v1
+#                                                  record (top-level shape)
 #   read_provenance(path)                -> dict | None  None on missing/corrupt
+#                                                  — including valid JSON whose
+#                                                  shape is not a v1 record
 #   batch_hash(paths, runner)            -> dict[str, str]  one git hash-object --stdin-paths
 #                                                            call; input-order-preserving; {}
 #                                                            on empty paths or runner failure
@@ -63,7 +67,12 @@
 #                                                  (source renamed/deleted — AC-05/T-06).
 #                                                  Both kinds filtered to drift_trigger:true only
 #                                                  (AC-09/T-05). Docs with no drift omitted (→ {}).
-#   load_provenance_dir(provenance_dir)  -> dict   {doc: record}; {} on missing/corrupt dir
+#   load_provenance_dir(provenance_dir, skipped=None)
+#                                        -> dict   {doc: record}; {} on missing/corrupt dir.
+#                                                  Per-FILE error boundary: one bad record
+#                                                  never suppresses a later valid one.
+#                                                  `skipped` (optional list) collects up to
+#                                                  MAX_SKIPPED_REPORTED rejected basenames.
 #   startup_drift_line(root, runner=None)
 #                                        -> str    "" when clean (AC-06);
 #                                                  "context drift: N stale source(s) across M doc(s) -- run /ca:context-check"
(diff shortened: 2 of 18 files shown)


SUaDtL:
Superseded by #436.

This branch was cut from `4ec1a8a`, before #427 merged. #427 shipped **ca-pi 0.1.3**, colliding with the version this PR claimed — the Pi release guard would reject it against the new base.

Re-opened as #436, rebased onto the current `main` and releasing **0.1.4** instead, with #427's 0.1.3 changelog section preserved intact. Same three fixes (#396, #413, #410), same 18-file diff.

A new PR rather than a force-push or a merge into this branch, deliberately: H-02 prohibits force-push, and merging `main` in would make the merge commit re-present #427's SHA-256 correlation lines as newly added crypto, tripping H-09b and asking for a crypto-gate pass on another PR's already-reviewed code.

Issue #428, opened 2026-07-25
Windows ca-pi adapter suites flake on hard 5s timing budgets


SUaDtL:
## Symptom

The `[CHECK] | [PI] | Adapter contract <os: windows-latest ...>` cells intermittently fail on timing, not logic. Two distinct instances observed during the 2026-07-24 backlog sweep:

1. **`test/background-jobs.test.ts > session-local background job state > launches and disposes a real Git Bash background job`**
   ```
   Error: Test timed out in 5000ms.
   × launches and disposes a real Git Bash background job  5015ms
   ```
   Missed the budget by **15ms**. Observed on PR #427, `windows-latest / Pi 0.80.10`. The other five matrix cells — including `windows-latest / Pi 0.80.5`, i.e. the same OS and the same code — all passed. The Pi runtime version is irrelevant to spawning Git Bash, so this is runner contention, not a behavioral difference.

2. **`test/process-tree.test.ts`** — `Windows Job Object holder refused containment: ready-timeout`, after the test's own built-in retry. Observed on run 30130392781, `windows-latest / Pi 0.80.5`.

## Why it matters

These are **required** merge-gate cells, so a flake blocks the PR and the cascading `[GATE] | [REPO] | Merge readiness` with it. It already cost one re-run this sweep and will keep taxing every ca-pi PR. Worse, a real Windows regression is now indistinguishable at a glance from noise — which is exactly how a genuine defect gets waved through as "just the flake".

## Root cause

Both tests spawn **real** OS processes (Git Bash; a Job Object holder) and assert against a fixed wall-clock budget. Windows GitHub runners are slower and noisier at process creation than Linux/macOS, so a budget tuned on a developer machine sits right at the edge.

## Suggested direction

Not simply "raise the timeout" — that trades a flake for a slow suite and hides real regressions. Prefer, in order:

1. Make the wait **condition-based** rather than duration-based where possible: poll for the observable readiness signal with a generous ceiling, so a fast runner stays fast and a slow one still passes.
2. Where a hard bound is genuinely part of the contract, give the Windows path an explicitly platform-scoped budget with a comment stating the measured baseline it derives from — not an arbitrary round number.
3. If a test's purpose is the *containment behavior* rather than its latency, decouple the assertion from wall-clock entirely.

## Acceptance criteria

- AC-1: Neither test asserts a bare fixed millisecond budget against real process spawn on Windows.
[...]

Issue #400, opened 2026-07-22
Upgrade vulnerable docs-site build dependencies


SUaDtL:
# Upgrade vulnerable docs-site build dependencies

**Severity:** medium | **Confidence:** 0.97 | **Effort:** M

**Where:**
- site/package-lock.json:2861-2864
- site/package-lock.json:3358-3363
- site/package-lock.json:7237-7242
- site/package-lock.json:7458-7462

**Evidence:** A live `npm audit --omit=dev --json` against `site/package-lock.json` reports three high-severity dependency findings: Astro 7.0.3 is within multiple affected ranges, direct Sharp 0.33.5 and Astro's Sharp 0.34.5 are below the fixed 0.35.0 line for GHSA-f88m-g3jw-g9cj, and transitive SVGO 4.0.1 is below 4.0.2 for GHSA-2p49-hgcm-8545. The lockfile locations above prove the exact vulnerable versions currently resolved.

**Impact:** The documentation build consumes versions with known XSS and image-processing vulnerabilities. Malicious or unexpectedly untrusted SVG/image/content input can execute unsafe build-time behavior or produce executable output, and the current production-dependency audit is red at high severity.

**Recommendation:** Update Astro/Starlight and Sharp to compatible fixed releases, ensure the resulting graph resolves SVGO 4.0.2 or newer and every Sharp copy to 0.35.0 or newer, then rebuild and verify the static site and lockfile policy.

**Acceptance criteria:**
- `npm --prefix site audit --omit=dev --audit-level=high` exits zero with no high or critical advisories.
- `site/package-lock.json` resolves Astro above 7.0.9, SVGO 4.0.2 or newer, and no Sharp version below 0.35.0.
- The site build, test, typecheck, and link audit pass against the regenerated lockfile.

Issue #390, opened 2026-07-22
Factor host-independent Pi checks out of the six-cell platform matrix


SUaDtL:
# Factor host-independent Pi checks out of the six-cell platform matrix

**Severity:** medium | **Confidence:** 0.99 | **Effort:** M

**Where:**
- .github/workflows/ci.yml:259-325

**Evidence:** The `ca-pi-tools` matrix expands three operating systems by two supported Pi versions, and every one of its six cells repeats package-metadata checking, TypeScript typecheck, the same package/activation/commands/status/bridge/tool-guard Vitest suites, static security/parity scripts, bundle rebuild and stale-output check, package tests, and a 100-sample benchmark before reaching the version/platform contract. In successful hosted run 29870271440, the six cells took 120, 120, 153, 188, 419, and 423 seconds: 1,423 runner-seconds (23.7 runner-minutes) for one PR run. Most repeated steps consume neither `matrix.os` nor `matrix.pi-version`; only external-host installation and live platform/version contracts need the full cross-product.

**Impact:** Pi-touching changes pay six installations and six copies of host-independent typecheck/build/static/unit work. The latest observed run spent 23.7 runner-minutes in this matrix and made its Windows cells the roughly seven-minute critical path, increasing queue pressure and feedback latency without adding six distinct verdicts for those invariant checks.

**Recommendation:** Split one canonical host-independent Pi job (typecheck, deterministic unit suites, security/parity scripts, build/stale check, static package checks, benchmark) from a lean six-cell OS/version compatibility matrix. Keep every test whose behavior genuinely depends on the installed Pi version or operating system in the matrix, and make the aggregate gate require both jobs.

**Acceptance criteria:**
- Host-independent Pi typecheck, build, static scripts, deterministic unit suites, and benchmark execute once per workflow run rather than once per OS/version cell.
- The supported Pi 0.80.5/0.80.10 by Windows/macOS/Linux compatibility matrix remains required and exercises every OS/version-sensitive live contract.
- The aggregate `ci-passed` gate requires both the canonical Pi job and all six compatibility cells.
- A hosted Pi-touching PR records lower total runner-seconds and no loss of required verdict coverage.

PR #432, opened 2026-07-25
fix(ci): add a pinned secret scan, gate the site deps, reach the parity fixture

### LINKED ISSUE #384: Run the Codex parity-fixture tests when the fixture generator changes
# Run the Codex parity-fixture tests when the fixture generator changes

**Where:**
- .github/workflows/ci.yml:17-42
- .github/workflows/ci.yml:119-125
- .github/workflows/ci.yml:430-435
- tools/codex-parity-fixture.py:1-181

**Evidence:** `test_codex_parity_fixture.py` runs only inside the path-scoped `hooks` job. That filter includes plugin hooks, core, `.github/scripts/**`, and the CI workflow, but not `tools/codex-parity-fixture.py`; the push trigger also omits the fixture. A pull request that edits only the 181-line fixture generator skips its own test, and the merged push starts no CI workflow for that path.

**Impact:** The live Claude/Codex parity scaffold can regress existing-target protection, bare onboarding shape, Git failure handling, or protected-artifact contents while CI remains green. Human parity sign-off may then rely on a malformed fixture and produce false enforcement evidence.

**Recommendation:** Add `tools/codex-parity-fixture.py` to the CI push paths and to the `hooks` filter (or give repo tools their own required job), and pin that wiring with a mutation-sensitive CI contract test.

**Acceptance criteria:**
- A pull request changing only `tools/codex-parity-fixture.py` runs `.github/scripts/test_codex_parity_fixture.py` as a required check.
- A push of that path to main also starts the relevant CI workflow.
- A structural CI test fails if either the source-path trigger or the test invocation is removed.

### LINKED ISSUE #403: Gate the docs-site dependency graph in CI
# Gate the docs-site dependency graph in CI

**Severity:** medium | **Confidence:** 0.99 | **Effort:** S

**Where:**
- .github/workflows/docs.yml:45-62
- .github/workflows/docs.yml:65-88
- .github/workflows/ci.yml:187-245
- .codearbiter/tech-stack.md:154-162

**Evidence:** The complete docs workflow installs `site/package-lock.json` twice and then runs typecheck/test/build/link-audit, but never runs `npm audit`. The repository CI workflow's only audit steps are scoped to the farm and sandbox workspaces. A live site audit currently reports three high advisories, demonstrating that a green docs workflow does not enforce the repository's dependency-risk posture.

**Impact:** Known vulnerable production/build dependencies can continue building and deploying to GitHub Pages with green CI. After a one-time upgrade, future advisories or lockfile regressions remain invisible until a manual audit or tribunal.

**Recommendation:** Add one required, lockfile-scoped audit step to the docs validation workflow after `npm ci`, with an explicit severity policy appropriate to the public site; keep its result in the deploy dependency chain so vulnerable builds cannot publish.

**Acceptance criteria:**
- The docs workflow runs `npm audit --omit=dev` at a documented severity threshold against `site/package-lock.json`.
- The deploy job cannot run when that audit step fails.
- A workflow contract test proves removing or weakening the site audit is detected.

### LINKED ISSUE #404: Add an independent repository secret scan to CI
# Add an independent repository secret scan to CI

**Severity:** medium | **Confidence:** 0.95 | **Effort:** M

**Where:**
- .codearbiter/tech-stack.md:162-173
- .github/workflows/ci.yml:1-881
- .github/workflows/docs.yml:1-104
- .github/workflows/release.yml:1-219

**Evidence:** The canonical tech-stack document explicitly states that no dedicated secrets scanner is configured; its two layers are a manual staged-diff sweep and local H-10b hook enforcement. A whole-unit read of every workflow confirms no independent secret-history or pull-request scan exists. Contributors, bots, or direct Git operations that do not run codeArbiter's cooperative local hooks therefore reach hosted CI without this check.

**Impact:** A real credential committed by a path outside the cooperative local gate can be merged and retained in Git history while all hosted checks stay green. The risk includes populated `.env.example` files, generated payloads, and contribution paths that never invoke H-10b.

**Recommendation:** Add a full-history-aware secret scanner as a pinned, read-only required CI job, with an explicit reviewed allowlist for the adversarial fixture tokens already present. Keep H-10b as the fast local gate; the hosted scan is an independent backstop, not a replacement.

**Acceptance criteria:**
- Pull requests run a pinned secret scanner over the relevant commit range and fail on an unallowlisted credential.
- Pushes to the default branch receive an equivalent scan or branch protection requires the pull-request check.
- Known fake keys in the secret-detection corpus are narrowly allowlisted by exact path/value rationale, and a new populated `.env.example` regression fails the check.

SUaDtL:
Three CI supply-chain gaps, one lane.

Closes #404
Closes #384
Closes #403

> **Revision 3.** Revision 2's adversarial review found two BLOCKs, one HIGH and two MEDIUMs — see [Round 2](#round-2--what-the-review-caught). A recheck of revision 2 then found the *narrowness contract itself* was bypassable: three BLOCKs, each measured guard-GREEN while the pinned scanner swallowed a live planted key. Auditing that defect class turned up eight more. All eleven are closed and re-measured — see [Round 3](#round-3--the-narrowness-contract-was-itself-bypassable). Every claim in this body is measured against the pinned image, not reasoned about.

---

### #404 — an independent, pinned repository secret scan

The repo had **no** hosted secret scanner, and both existing layers are **cooperative**: H-10b blocks a commit whose diff touches a secret line, and `/ca:preview` sweeps the working diff — but both only run inside a contributor's local Claude/Codex/Pi session. A bot, a fork, a github.com web edit, or a plain `git push` runs neither.

New `secret-scan` job in `ci.yml`:

- **Pinned by image digest**, not a tag: `ghcr.io/gitleaks/gitleaks@sha256:c00b6bd0…abbb7f` (v8.30.1). A scanner runs over the entire repository — exactly the thing that must not be swappable by whoever controls the upstream tag.
- **Read-only**: `--network none --read-only --cap-drop ALL --security-opt no-new-privileges --user $(id -u):$(id -g)`, over a `:ro` mount.
- **Two scan modes** (see the HIGH finding below): `gitleaks dir` over the merge-result tree, plus `gitleaks git --log-opts <merge-base>..HEAD` over the PR's own commits.
- **Actionable output**: `--redact --verbose` — file, line, rule, commit; `Secret: REDACTED`.
- **Always-on**, no `needs: changes` gate, registered in **both** `ci-passed` registrations per the #390 contract.

> Not chosen: `gitleaks/gitleaks-action` — it requires a paid `GITLEAKS_LICENSE` for org accounts, and a digest-pinned container is a stronger pin anyway.

### #384 — the Codex parity fixture now reaches its own test

`.github/scripts/test_codex_parity_fixture.py` runs inside the path-scoped `hooks` job, but neither the push trigger nor the `hooks` paths-filter listed `tools/codex-parity-fixture.py`. A PR editing only that generator **skipped its own test**. Added to both.

### #403 — the docs-site dependency graph is gated

`docs.yml` ran `npm ci` → typecheck → test → build → link-audit and never `npm audit`. Added `npm audit --omit=dev --audit-level=high` to `site-check`; `deploy` already `needs: [build, site-check]`, so a red audit blocks the publish.

---

### BLOCK 1 — the allowlist guard never ran on a diff that weakens it

`test_the_secret_scan_allowlist_stays_narrow_and_value_scoped` lives in `test_ci_impact.py`, which runs **only** in the path-scoped `ci-impact` job. `.gitleaks.toml` matched no `changes` filter and was absent from the push trigger, so a PR whose only change was to widen the allowlist **skipped the one job that guards it** — the identical defect class as #384, which this very branch fixes elsewhere. `.github/workflows/docs.yml` had the same gap, which is also the MEDIUM against #403 AC-3: the diff shape that deletes the site audit step is exactly the one the guard could not see.

Fixed on three levels, because a path filter is a promise this repo has now broken twice:

1. `.gitleaks.toml` and `.github/workflows/docs.yml` added to the `impact` filter.
2. Both added to the push trigger.
3. **Belt and braces** — the narrowness contract now *also* runs inside the `secret-scan` job, which carries no `needs: changes` gate. No future filter edit can leave the allowlist unguarded.

**Proof, as asked — a diff whose only change is `.gitleaks.toml`:**

```
A. Would the `changes` job set impact=true?
   changed files : ['.gitleaks.toml']
   matching globs: ['.gitleaks.toml']
   => impact == 'true'? True   => the ci-impact job RUNS the guard

B. Is the guard also reachable with NO filter involvement?
   secret-scan job-level `needs:` present? False
   secret-scan job-level `if:`    present? False
   job invokes the narrowness guard?       True
   => unconditional: every PR reaching CI re-proves the allowlist

C. Does the guard BITE on the widening diff?
   exit=1
   FAIL: test_the_secret_scan_allowlist_stays_narrow_and_value_scoped
   FAIL: test_a_broad_secret_scan_waiver_is_rejected_by_the_narrowness_contract
   AssertionError: Lists differ: ['`regexTarget = "line"` binds the waiver ...'] != []
```

### BLOCK 2 — the contract enforced literalness, not narrowness

The old test required each waiver to match `\A[A-Za-z0-9_-]+\Z` and never constrained `regexTarget`. The review demonstrated that `regexTarget = "line"` turns a fixed literal into a substring match against every scanned line.

**The review's prescribed fix — "constrain `regexTarget`" — is necessary but NOT sufficient, and I want to flag that explicitly rather than quietly over-deliver.** gitleaks tests allowlist regexes with Go's `FindString`, a **substring** search. So even with `regexTarget` omitted (matching against the detected secret itself), an unanchored literal is as broad as its haystack. Measured against the pinned image over the tracked tree, with a high-entropy key planted in `plugins/ca/tools/.env.example`:

| allowlist shape | result |
| --- | --- |
| *(shipped config, key planted)* | **`leaks found: 1`, exit 1** |
| `regexTarget = "line"`, `regexes = API_KEY` | `no leaks found` |
| `regexTarget = "match"`, `regexes = API_KEY` | `no leaks found` |
| `regexTarget` omitted, `regexes = sk-` | `no leaks found` ← **the case constraining `regexTarget` misses** |
| `regexTarget` omitted, `regexes = \Ask-REPLACE_ME\z` | **`leaks found: 1`, exit 1** |

So the rule is now **both**: `regexTarget` is forbidden outright *and* every waiver must be anchored `\A<literal>\z` with no regex metacharacter in the body — true only when the **entire detected secret** is that exact fixture value.

The anchored values are the secrets gitleaks *actually reports*, derived by measurement (`--report-format json` against the bare default ruleset) rather than by reading the source: **13 findings across 5 distinct values**. That is also what makes the PEM waiver honest — the reported secret there is a five-line block of constant *names* (`KEY_BODY_1/2/3`), so the waiver is that exact block anchored, and any other PEM is a different secret. Reformatting the fixture turns the job red until the waiver is re-derived, which is the correct direction to fail.

The narrowness contract is now a pure function (`gitleaks_waiver_violations`) exercised against adversarial configs that are never written to disk, so `test_a_broad_secret_scan_waiver_is_rejected_by_the_narrowness_contract` proves the contract fails on every shape above — including the `line`-target waiver the review asked for by name.

### HIGH — #404 AC-1 asked for the commit RANGE

Correct, and undeclared. The job scanned `git archive HEAD` — the merge-result **tree** — and never inspected history. Not theoretical; measured on a scratch repo where a credential is added in commit 1 and deleted in commit 2:

```
===== TREE  scan (gitleaks dir .)            -> exit 0 =====
INF no leaks found

===== RANGE scan (gitleaks git --log-opts)   -> exit 1 =====
Finding:     aws_secret_access_key = "REDACTED"
Secret:      REDACTED
RuleID:      generic-api-key
File:        leak.py
Line:        1
Commit:      438b8cc50df157eca82f3bb1a57c6cba00cf5d96
WRN leaks found: 1
```

The repo allows merge and rebase merges, so those commits reach `main` verbatim. A second, `pull_request`-scoped step now scans the range. The workspace is mounted `:ro` **with** its `.git`; `safe.directory` is passed via `GIT_CONFIG_*` (nothing is writable) and `/tmp` is a small `noexec` tmpfs so git has scratch space without a writable rootfs. Verified under the full hardened flag set including `--user`.

### MEDIUM — a failing scan was unactionable

`--redact` without `-v` prints only `leaks found: N`, while the job comment claimed the file/line was available. `--verbose` restores `File:`/`Line:`/`RuleID:`/`Fingerprint:` (plus `Commit:` in git mode) and still prints `Secret: REDACTED` — confirmed in every scan output quoted in this body. The comment is rewritten to match what the flags actually do.

### LOW — the threshold tightening was inert and misattributed

Confirmed and corrected. Measured `dependencies` from each `package.json`:

| graph | production deps | audited by |
| --- | --- | --- |
| `plugins/ca/tools` | **0** | `ci.yml` — `tools` |
| `plugins/ca-sandbox/tools` | **0** | `ci.yml` — `ca-sandbox-tools` |
| `plugins/ca-pi/tools` | **0** | `ci.yml` — `ca-pi-checks` *(new)* |
| `site` | 3 (astro, starlight, markdown-remark) | `docs.yml` — `site-check` |

`--omit=dev` therefore audits an **empty graph** in three of four jobs, at any threshold. The comment claiming "every advisory cleared by #400 was rated HIGH, so the gate would have passed all of them" was true of the *site* gate and false of the farm step it annotated — #400's advisories were all in `site/package-lock.json`. Both `ci.yml` and `tech-stack.md` now describe the farm/sandbox/ca-pi thresholds as a **durability** setting and name the site gate as the part with immediate effect.

Two follow-ons the review surfaced:

- `plugins/ca-pi/tools` is a **published** host package that ran `npm ci` and never audited it, which made `tech-stack.md`'s "every dependency graph the repo ships or publishes" false. It is now gated, and a new assertion requires **every job that installs a lockfile to audit it**.
- The dev trees are where the real advisories are: measured in `plugins/ca/tools`, `npm audit --json` reports `{prod: 1, dev: 104}` and one **HIGH** (`postcss`) the shipped gate never sees. **Issue #434** tracks extending the sweep, and notes the blast radius — those dev deps build `farm.js` and `sandbox.js`, committed shipped artifacts.

---

## Red test evidence

All new assertions written and run **before** implementation — 9 failures across 7 test methods, each for the defect and not an import error:

```
FAIL: test_the_secret_scan_allowlist_stays_narrow_and_value_scoped
AssertionError: Lists differ: ['`regexTarget = "line"` binds the waiver [2415 chars] it"] != []
First list contains 18 additional elements.
 : the allowlist can waive more than the fixture values it names

FAIL: test_a_broad_secret_scan_waiver_is_rejected_by_the_narrowness_contract
AssertionError: Lists differ: [...] != [] : the shipped allowlist is not narrow

FAIL: test_secret_scan_config_edits_reach_the_guard_that_constrains_them (path='.gitleaks.toml')
AssertionError: '.gitleaks.toml' not found in ['.github/ci-impact-map.json',
  '.github/scripts/test_ci_impact.py', 'core/hosts.json', 'tools/ci-impact.py',
  'tools/host_descriptors.py', '.github/workflows/ci.yml']
 : a PR touching only this file skips its own contract test

FAIL: test_secret_scan_config_edits_reach_the_guard_that_constrains_them (path='.github/workflows/docs.yml')
AssertionError: '.github/workflows/docs.yml' not found in [...]

FAIL: test_the_allowlist_guard_also_runs_unconditionally_in_the_secret_scan_job
AssertionError: 'test_ci_impact.WorkflowContractTest.test_the_secret_scan_allowlist_stays_narrow_and_value_scoped'
  not found in '  secret-scan:\n    name: "[CHECK] | [REPO] | Secret scan"\n ...'
 : the always-on job does not re-prove its own allowlist

FAIL: test_the_secret_scan_covers_the_pull_requests_commit_range
AssertionError: Regex didn't match: '(?m)^\\s+git \\. \\\\$'
 : the secret scan never inspects commit history

FAIL: test_ci_runs_a_pinned_read_only_secret_scan_wired_into_the_merge_gate
AssertionError: 1 != 2 : expected exactly the tree scan and the commit-range scan

FAIL: test_every_npm_audit_gate_fails_the_build_on_a_high_advisory
AssertionError: 3 != 4 : expected exactly the farm, sandbox, ca-pi, and site audit gates

Ran 32 tests
FAILED (failures=9)
```

### The contracts BITE — 14/14 mutations caught

Each fix reverted against a mutated copy of its subject file, then restored byte-identically:

```
BITES   BLOCK1 impact filter drops .gitleaks.toml
BITES   BLOCK1 impact filter drops docs.yml
BITES   BLOCK1 push trigger drops .gitleaks.toml
BITES   BLOCK1 secret-scan stops re-proving the allowlist
BITES   BLOCK1 secret-scan becomes path-scoped (needs: changes)
BITES   BLOCK2 a waiver loses its anchors
BITES   BLOCK2 regexTarget = "line" comes back
BITES   BLOCK2 a wildcard is smuggled inside the anchors
BITES   BLOCK2 a paths waiver returns
BITES   HIGH  the commit-range scan is deleted
BITES   HIGH  the range scan loses its --log-opts scoping
BITES   HIGH  the checkout goes shallow again
BITES   MED   the tree scan loses --verbose (unactionable log)
BITES   LOW   the ca-pi audit gate is dropped

14/14 mutations caught
```

Two mutations initially reported MISSED and were **my harness's fault, not a contract gap** — `fetch-depth: 0` appears 5× in `ci.yml` so the mutation hit the wrong job, and the ca-pi mutation renamed the step while leaving its `run:` line intact. Both re-anchored; both then bit.

### The scanner itself was run, not just configured

Against the current merge result:

```
1. shipped anchored config, unmodified tree:
exit=0  no leaks found

2. high-entropy key planted in plugins/ca/tools/.env.example (#404 AC-3):
exit=1  leaks found: 1
      Secret:      REDACTED
      RuleID:      generic-api-key
      File:        plugins/ca/tools/.env.example
      Line:        4

3. real credential planted INSIDE an allowlisted fixture file:
exit=1  leaks found: 1
      Secret:      REDACTED
      RuleID:      generic-api-key
      File:        .github/scripts/test_hooklib.py
      Line:        760
```

Case 3 is the one that matters for the allowlist design: an allowlisted fixture file is still **fully scanned**. (A first attempt at this probe reported a false pass; the planted line used escaped quotes that gitleaks does not match **even with no allowlist at all**, so it was a bad probe rather than a blind spot. Re-run with a well-formed assignment, it fires.)

---

## Suites

| suite | result |
| --- | --- |
| `.github/scripts` | **1074 tests**, 4 failures — all `test_pi_benchmark` ("requires the installed pinned esbuild"), environmental and untouched by this diff |
| `plugins/ca/hooks/tests` | **1057 OK** |
| `tools/sync-core.py --check` | OK (47 core files × 3 plugins, byte-identical) |
| `tools/build-surface.py --check` | OK (claude, codex, pi in sync) |
| payload gate | no Pi payload change — version bump not required |
| `git diff --check origin/main HEAD` | exit 0 |
| `.gitleaks.toml` TOML parse | OK (5 allowlists, `regexTarget` omitted in all) |
| `ci.yml` / `docs.yml` YAML parse | OK (24 / 3 jobs) |

`origin/main` advanced 4 commits during this round and was merged in (not rebased); every measurement above is against the merge result.

**Live CI evidence** — the new `secret-scan` steps running for real (run [30145274733](https://github.com/arbiterForge/codeArbiter/actions/runs/30145274733)), not skipped:

```
test_the_secret_scan_allowlist_stays_narrow_and_value_scoped ... ok
test_a_broad_secret_scan_waiver_is_rejected_by_the_narrowness_contract ... ok
Ran 2 tests in 0.007s
OK
INF scanned ~13737893 bytes (13.74 MB) in 6.69s
INF no leaks found
scanning commit range 9e8768fcf79dc76378ab03cb45bbcfc2c48d212b..HEAD
INF 2 commits scanned.
INF scanned ~54382 bytes (54.38 KB) in 48.9ms
INF no leaks found
```

**Correction to revision 1's NEEDS-TRIAGE #1:** it reported 7 pre-existing failures in `plugins/ca/hooks/tests`. That suite is now **1057 OK** — those failures are gone as of current `origin/main`, so the item is withdrawn rather than carried forward.

**H-10b / H-09b gate:** this diff touches credential-shaped literals, so the commit hook correctly blocked it. The gate was then genuinely worked: every waiver literal is a **re-anchoring of a value already in this same file** on `origin/main` — no new credential material. The only new text is the PEM waiver, which carries BEGIN/END *marker* text plus three test-local *constant names*; the fake key bodies are not reproduced. No cryptographic operation is introduced (`BEGIN RSA PRIVATE KEY` appears only as a literal inside a scanner regex; the `sha256:` occurrences are image digest pins). The sole sink is gitleaks' matcher inside a network-isolated, read-only container running `--redact`. Net direction is strictly narrower, which the three measurements above confirm.

---

## Round 3 — the narrowness contract was itself bypassable

Revision 2 added a contract asserting the `.gitleaks.toml` allowlist cannot be silently widened. A recheck proved **the contract could be walked past**: three BLOCKs, each measured *guard GREEN while the pinned scanner swallowed a live planted key*.

Auditing the same defect class while fixing them turned up **eight more**. All eleven are one bug: **the contract read `.gitleaks.toml` as TEXT**, and TOML — plus viper's case-insensitive key lookup — admits far more spellings than a regex anticipates.

### How every shape was measured

```
tree     git archive HEAD (5ad8964), plus a planted key in plugins/ca/tools/.env.example:
         FARM_API_KEY=sk-kR3mVq7XpZ2wNbT9sLdG4hJfA6cE1yUo
scanner  ghcr.io/gitleaks/gitleaks@sha256:c00b6bd0…abbb7f   (v8.30.1 — the pinned digest)
command  gitleaks dir . --config /repo/.gitleaks.toml --redact --exit-code 1
```

**Baseline — the shipped config sees the planted key:**

```
INF scanned ~13737942 bytes (13.74 MB) in 2.21s
WRN leaks found: 1
exit=1
```

### The three reported BLOCKs, verbatim

**1. Indented `regexes` key.** `gitleaks_allowlist_regexes()` extracted with `(?ms)^regexes\s*=\s*\[` — anchored at column 0 — while "names no value" was the substring test `"regexes = [" in block`. Valid TOML with leading whitespace therefore yielded **zero** literals *and* still read as bounded.

```toml
[[allowlists]]
description = "adversarial: indented regexes key"
  regexes = ['''sk-''']
```
```
INF scanned ~13737942 bytes (13.74 MB) in 2.18s
INF no leaks found
exit=0
```

**2. `disabledRules` under `[extend]`.** The contract asserted `useDefault = true` was present and never looked at what sat beside it. This deletes the highest-value rule outright.

```toml
[extend]
useDefault = true
disabledRules = ["generic-api-key"]
```
```
INF scanned ~13737942 bytes (13.74 MB) in 2.22s
INF no leaks found
exit=0
```

**3. Single-quoted `regexTarget`.** The ban matched double quotes only (`regexTarget\s*=\s*"([^"]*)"`). Bound to the `match` target, a whole-match-anchored literal also clears the anchoring *and* metacharacter checks.

```toml
[[allowlists]]
description = "adversarial: single-quoted regexTarget"
regexTarget = 'match'
regexes = ['''\AFARM_API_KEY=sk-kR3mVq7XpZ2wNbT9sLdG4hJfA6cE1yUo\z''']
```
```
INF scanned ~13738005 bytes (13.74 MB) in 2.18s
INF no leaks found
exit=0
```

> Worth recording: the same waiver bound to `regexTarget = 'line'` does **not** swallow the key, because gitleaks' `line` target carries a **leading newline**. `\AFARM_API_KEY=…\z` misses it; `'match'` is the shape that lands. The contract now rejects the key outright rather than reasoning about which target is survivable.

### Eight more of the same class, found while fixing those three

| # | shape | measured effect |
|---|---|---|
| 4 | `regexes = ["sk-"]` (basic string) | only `'''…'''` was ever extracted → zero literals |
| 5 | `regexes = ['sk-']` (literal string) | same |
| 6 | indented `[[allowlists]]` header | column-0 block scan → an entire block invisible |
| 7 | `RegexTarget = "match"` | viper's key lookup is case-**in**sensitive; the bans were not |
| 8 | `Paths = ['''.''']` | **`scanned ~0 bytes (0) in 1.46ms`** — the whole repo went unread |
| 9 | `DISABLEDRULES = [...]` | same case hole, applied to the `[extend]` ban |
| 10 | `]` inside a waiver literal | array capture was non-greedy to the first `]`; every later waiver in that array went unchecked |
| 11 | `[[rules]]` re-declaring `generic-api-key` | a matching id **replaces** the default rule — and the contract never looked at `[[rules]]` at all |

Shape 8, verbatim — a single capitalised letter un-scans the repository:

```
INF scanned ~0 bytes (0) in 1.46ms
INF no leaks found
exit=0
```

### The before/after asymmetry

`gitleaks_waiver_violations` as it shipped at `5ad8964` vs. at HEAD, run over the identical adversarial configs:

```
adversarial .gitleaks.toml                     BEFORE       AFTER
------------------------------------------------------------------
indented `regexes` key                         PASSES    rejected
`disabledRules` under [extend]                 PASSES    rejected
single-quoted `regexTarget = 'match'`          PASSES    rejected
`regexes = ["sk-"]` (basic string)             PASSES    rejected
`regexes = ['sk-']` (literal string)           PASSES    rejected
indented `[[allowlists]]` header               PASSES    rejected
`RegexTarget = "match"` (case)                 PASSES    rejected
`Paths = ['''.''']` (case)                     PASSES    rejected
`DISABLEDRULES` (case)                         PASSES    rejected
`]` inside a waiver literal                    PASSES    rejected
`[[rules]]` re-declaring generic-api-key       PASSES    rejected
------------------------------------------------------------------
THE SHIPPED CONFIG (13 waivers, 5 blocks)      PASSES      PASSES

shipped waivers extracted: 13 across 5 blocks
```

Every shape in that table is one that returned the planted key to `no leaks found`, exit 0. The legitimate shipped config — 13 anchored, metacharacter-free waivers across 5 blocks — still passes, and still reports `leaks found: 1` (exit 1) on the planted key.

### The fix — parse it, don't scrape it

`gitleaks_waiver_violations` now parses the document with `tomllib` and folds keys to lower case, the way gitleaks' own viper-backed reader does. On top of that:

- **Default-deny on waiver keys** — a waiver may carry only `description` and `regexes`. `paths`/`commits`/`stopwords`/`regexTarget` keep their specific messages; anything else is rejected by name. The next widening knob gitleaks ships fails this contract the day it is typed, not the day someone remembers to ban it.
- **`[extend]` may set nothing but `useDefault`.**
- **`[[rules]]` is rejected** — this file only ever *subtracts* fixture values.
- **Unparseable TOML is a violation**, not a silent pass. Reporting "narrow" on a config that cannot load anything is the worst possible failure direction.

`.gitleaks.toml`'s own "ADDING AN ENTRY" rules were rewritten to match, including an explicit note that reformatting will not help.

### Audited and found clean

Same class, checked against the pinned image, **not** exploitable:

| shape | why it is safe |
|---|---|
| inline `allowlists = [{…}]` | gitleaks **ignores** it — `leaks found: 1` |
| singular `[allowlist]` | `FTL [allowlist] is deprecated, it cannot be used alongside [[allowlists]]` |
| `[extend] path = …` | `FTL unable to load config due to extend.path and extend.useDefault being set` |
| `[extend] url = …` | the scan job runs `--network none`; nothing can be fetched — `leaks found: 1` |

The inline-array-of-tables case is still guarded: the contract asserts the parsed allowlist count equals the table-header count, so a waiver the scanner ignores and a reader believes in fails the build.

---

### MEDIUM — `docs.yml` never ran on itself

Its triggers were `site/**` and `plugins/ca/**` only. **The commit that added the site audit step did not execute it** — 37 checks reported on this PR and not one `Site |` among them. A PR deleting that audit, or breaking the `deploy → site-check` edge, touches `docs.yml` and nothing else: exactly the diff its own jobs could not see. This is the same defect class as #384 and the `.gitleaks.toml` gap this branch already fixes.

`.github/workflows/docs.yml` added to both its own `push` and `pull_request` path filters, with a test asserting it. Verified the newly reachable gate is green before making it reachable:

```
$ cd site && npm audit --omit=dev --audit-level=high
found 0 vulnerabilities
```

### LOW — the npm-audit sweep over-claimed; **derived** rather than narrowed

`test_every_npm_audit_gate_fails_the_build_on_a_high_advisory` said *"every job that installs a lockfile must also audit it"* while iterating the hardcoded triple `("tools", "ca-sandbox-tools", "ca-pi-checks")`. The claim was already false: `ca-pi-tools` is a **required** gate that runs `npm ci` and never audits — and so does `docs.yml`'s `build`. Both are benign only because a sibling job audits the same lockfile, and nothing checked that.

**Chose derivation, not a narrower comment.** A comment that matches three hardcoded names still cannot see a fourth graph, and that is precisely how `site/package-lock.json` went unguarded through #400/#403 — the exact failure this lane exists to close. Narrowing the prose would have made the test honest and left the hole open.

The rule is now the one that actually matters: **a job may skip the audit only when another job in the same workflow audits the very directory it installs.** No job list appears in the test. It is proven to bite by a synthetic job installing a brand-new graph, which the check rejects.

Two extraction bugs surfaced while deriving it — both the same read-it-as-text class as the allowlist fix, both fixed:

- the trigger/filter glob readers stopped at the first entry carrying a **trailing comment**, so `docs.yml`'s push filter read as `['site/**']` and `plugins/ca/**` was invisible to every assertion over it;
- the `npm ci` / `npm audit` detectors matched only the `- name:`/`run:` step shape, so a job using the inline `- run: npm ci` spelling would have gone entirely unseen.

### The gate caught this branch

Worth recording, because it is the whole thesis of the lane demonstrated on its own author. The first push of Round 3 turned `[CHECK] | [REPO] | Secret scan` **red**:

```
RuleID:      generic-api-key
File:        .github/scripts/test_ci_impact.py
Line:        1349
leaks found: 1
```

The commit had pasted the real high-entropy key used to *measure* the `regexTarget` bypass into a tracked file. The scanner was right. Fixed by removing the key, **not** by waiving it — widening the allowlist to admit a credential this branch itself introduced would invert the entire point of the branch. The value is now a low-entropy stand-in, which costs the assertion nothing: `regexTarget` is rejected on sight, whatever literal it is paired with.

Reproduced and re-verified locally against the pinned image over the real tracked tree, the way the CI job scans it:

```
before   RuleID: generic-api-key / test_ci_impact.py:1349 / leaks found: 1   exit 1
after    scanned ~13759566 bytes (13.76 MB) / no leaks found                 exit 0
```

### OPEN BLOCKER — the key is still in this branch's HISTORY

The working tree is clean, but `fc55acd` **added** the key and `8ec6e6d` removed it. The tree scan is green; the commit-range scan is not, and that is the range scan doing the exact job it was added for:

```
$ gitleaks git . --log-opts a3a2df2..HEAD --config .gitleaks.toml --redact --verbose
RuleID:      generic-api-key
File:        .github/scripts/test_ci_impact.py
Line:        1228
Commit:      fc55acd910a513c1eb14a60582614e3b9404a0f6
INF 14 commits scanned.
WRN leaks found: 1
```

Current CI on `8ec6e6d`: **37 pass, 2 fail** — `Secret scan` (the above) and `Merge readiness` (its consequence). The Windows Pi adapter flake (#428) passed this run.

**Deliberately not resolved unilaterally.** The three available routes each break something this lane is built on, so the choice belongs to the maintainer:

1. **Rewrite `fc55acd`/`15d8cdf` to drop the key and force-push.** The only thing that actually removes a blob from history. Requires force-pushing a pushed PR branch, which this lane's rules prohibit without explicit authorization. *(Recommended — this is a solo branch with no downstream clones, and history is the only place the credential still exists.)*
2. **Abandon this branch and re-open from a fresh one** built as a single clean commit off `origin/main`. Never rewrites a pushed branch, but changes the PR number.
3. **Add a `.gitleaksignore` fingerprint** pinned to commit+file+rule+line. Narrow in itself, but it introduces a waiver surface the narrowness contract does **not** police — i.e. a twelfth bypass of exactly the class this PR just closed eleven of. It would need the contract extended to cover `.gitleaksignore` first, and even then it waives a credential this branch itself introduced.

Widening `.gitleaks.toml`'s allowlist is not on the list, and will not be.

### Test state

```
$ cd .github/scripts && python -m unittest discover -s . -p "test_*.py"
Ran 1075 tests in 70.011s
FAILED (failures=2, errors=2, skipped=5)
```

All four are `test_pi_benchmark`, all `RuntimeError: Pi benchmark requires the installed pinned esbuild` — environmental, a fresh worktree with no `node_modules`, unrelated to this diff. `test_ci_impact` alone: **33 tests, OK.**

---

## NEEDS-TRIAGE — not in this diff

1. **`site/src/content/docs/guides/adding-a-dependency.md:62-65`** still documents the CVE gate as `--audit-level=critical`. That describes the `/ca:add-dep` skill's gate rather than CI's, but the repo now documents two thresholds. Fixing it means touching `site/**`, which requires the full `npm test` + build + link-audit loop; deliberately out of a CI-scoped lane.

2. **`docs.yml`'s `site-check` is not a required status check.** Branch protection on `main` requires only `[GATE ] | [REPO] | Merge readiness` from `ci.yml`, so a red site audit blocks the Pages publish (#403 AC-2, satisfied) but not the merge itself. Raised by the review as pre-existing; it is a branch-protection setting, not a diff.

3. **codeArbiter's own per-line redactor degrades `/ca:doctor` when the project path contains a trigger word.** Found while verifying the suite: in a worktree at `…/pr432-secret-scan-narrowness`, `test_pi_package.test_real_isolated_rpc_command_discovery_and_keyed_status` fails because the `package`, `core` and `child` diagnosis lines — the three that embed the project path — are replaced by `[REDACTED — secret-pattern match removed before transmission]`. Confirmed by control: the same commit passes in worktrees whose paths lack the word, and a purpose-named `secret-probe` worktree at the same commit reproduces the failure. Environmental, not this diff, but a real papercut — a benign path containing `secret` silently degrades doctor output.

4. **`ci.yml`'s push trigger is path-filtered**, so a push to `main` touching only unlisted paths starts no workflow and therefore no secret scan. #404 AC-2 is satisfied through the PR check, which is `main`'s enforcement point. Broadening the push paths would drag every other job along.

---

## CI state — the red gate is pre-existing, tracked as #428

> Describes **revision 2's** run; revision 3's CI is in flight. The #428 analysis below is unaffected by it.

35/37 checks pass. The two that do not are **not this lane's**:

```
[CHECK] | [PI  ] | Adapter contract <os: windows-latest · runtime: Pi 0.80.10>  -> fail
[GATE ] | [REPO] | Merge readiness                                             -> fail  (consequence of the above)
```

The failure is `test/background-jobs.test.ts > launches and disposes a real Git Bash background job`, `Error: Test timed out in 5000ms` — a real-process spawn against a hard 5s budget on a Windows runner.

**Reproduced on `origin/main`, not introduced here.** Run [30138368211](https://github.com/arbiterForge/codeArbiter/actions/runs/30138368211) on `main` @ `9e8768fc` — the exact commit this branch merged — fails on the *same test with the same message*, differing only in which Pi matrix cell it lands in (0.80.5 there, 0.80.10 here), which is itself the signature of a timing flake rather than a version-specific defect. Re-running the failed jobs reproduced it.

This diff touches **no Pi file** (`git diff --name-only origin/main HEAD` → `tech-stack.md`, `test_ci_impact.py`, `ci.yml`, `docs.yml`, `.gitleaks.toml`). Already tracked as **#428 — "Windows ca-pi adapter suites flake on hard 5s timing budgets"**; out of lane scope and deliberately not papered over here.

Every check this lane owns is green, including the new work: `[CHECK] | [REPO] | Secret scan` passed with all seven steps executing — the allowlist guard, the tree scan, and the commit-range scan.

### DIFF
--- a/.codearbiter/tech-stack.md
+++ b/.codearbiter/tech-stack.md
@@ -153,23 +153,41 @@ node -e "JSON.parse(require('fs').readFileSync(process.argv[1],'utf8'))" <file>.
 
 ## CVE gate (supply chain)
 
-The configured CVE gate is `npm audit --omit=dev --audit-level=critical`, run in
-CI (the `tools` job, after `npm ci`) against the shipped dependency set. A
-CRITICAL advisory fails the build; lower severities are intentionally not
-gating, so routine dev-tool advisories don't block unrelated PRs. This enforces
-the supply-chain posture described in `security-controls.md`.
+The configured CVE gate is `npm audit --omit=dev --audit-level=high`, run after
+`npm ci` against every dependency graph the repo ships or publishes:
+
+| graph | workflow / job |
+| --- | --- |
+| `plugins/ca/tools` | `ci.yml` — `tools` |
+| `plugins/ca-sandbox/tools` | `ci.yml` — `ca-sandbox-tools` |
+| `site` (docs site) | `docs.yml` — `site-check`, which `deploy` needs |
+
+A HIGH-or-worse advisory fails the build; `--omit=dev` scopes the gate to
+shipped dependencies so routine dev-tool advisories don't block unrelated PRs.
+The threshold was `critical` until issue #403: every advisory cleared by #400
+was rated HIGH, so the gate would have passed all of them. This enforces the
+supply-chain posture described in `security-controls.md`.
 
 ## Secrets scan
 
-No dedicated secrets scanner is configured. The gate is two-layered:
+The gate is three-layered — one hosted backstop plus two cooperative local
+layers:
 
-1. Manual sweep of the staged diff for credential patterns
+1. `ci.yml`'s `secret-scan` job (issue #404): gitleaks' full default ruleset,
+   run over the tracked tree on every pull request as a required check for the
+   `ci-passed` merge gate. The scanner is pinned by image digest and runs in a
+   network-isolated, read-only container over a read-only mount. This is the
+   only layer a bot, a fork, a web edit, or a plain `git push` cannot skip — the
+   two below run only inside a contributor's local session. Its allowlist
+   (`.gitleaks.toml`) waives individual fake fixture literals by value, never a
+   file path: a `paths` waiver would drop the whole file from the scan.
+2. Manual sweep of the staged diff for credential patterns
    (`api[_-]?key|token|secret|password|private[_-]?key|passphrase|credential|BEGIN.*PRIVATE|AKIA|ghp_|sk-ant`),
    case-insensitive. This is the convenience layer; the authoritative classifier
    is `_hooklib.SECRET_RE`, pinned against the farm redactor by the shared
    `hooks/secret-detection-corpus.json`, which also matches a secret keyword as
    the trailing segment of a compound name (e.g. `FARM_API_KEY = "..."`).
-2. The plugin's own enforcement hooks: H-09b (crypto/TLS) and H-10b (secrets)
+3. The plugin's own enforcement hooks: H-09b (crypto/TLS) and H-10b (secrets)
    block any commit whose diff touches a crypto or secret line until a diff-bound
    security-gate pass marker covers those exact lines (`hooks/security-pass.py`).
 
--- a/.codearbiter/tech-stack.md
+++ b/.codearbiter/tech-stack.md
@@ -156,17 +156,36 @@ node -e "JSON.parse(require('fs').readFileSync(process.argv[1],'utf8'))" <file>.
 The configured CVE gate is `npm audit --omit=dev --audit-level=high`, run after
 `npm ci` against every dependency graph the repo ships or publishes:
 
-| graph | workflow / job |
-| --- | --- |
-| `plugins/ca/tools` | `ci.yml` — `tools` |
-| `plugins/ca-sandbox/tools` | `ci.yml` — `ca-sandbox-tools` |
-| `site` (docs site) | `docs.yml` — `site-check`, which `deploy` needs |
+| graph | workflow / job | production deps today |
+| --- | --- | --- |
+| `plugins/ca/tools` | `ci.yml` — `tools` | none |
+| `plugins/ca-sandbox/tools` | `ci.yml` — `ca-sandbox-tools` | none |
+| `plugins/ca-pi/tools` | `ci.yml` — `ca-pi-checks` | none |
+| `site` (docs site) | `docs.yml` — `site-check`, which `deploy` needs | astro, starlight, markdown-remark |
 
 A HIGH-or-worse advisory fails the build; `--omit=dev` scopes the gate to
 shipped dependencies so routine dev-tool advisories don't block unrelated PRs.
-The threshold was `critical` until issue #403: every advisory cleared by #400
-was rated HIGH, so the gate would have passed all of them. This enforces the
-supply-chain posture described in `security-controls.md`.
+This enforces the supply-chain posture described in `security-controls.md`.
+
+**What the threshold actually buys, honestly.** Three of the four graphs above
+declare no production dependencies at all, so `--omit=dev` audits an empty graph
+in those jobs and the threshold there is inert at any value — it is a
+*durability* setting that decides what happens the day one of them takes on a
+runtime dependency. The one live graph is `site`, and until issue #403 nothing
+audited it: #400's three HIGH advisories all lived in `site/package-lock.json`,
+which is why the site gate is the part of this change with immediate effect. The
+threshold was lowered from `critical` to `high` across all four so a single rule
+covers the repo — not because the farm or sandbox audits would have caught #400.
+They would not have seen it.
+
+The dev trees, where these packages' advisories actually are, are **not** covered
+by this gate. Measured in `plugins/ca/tools` at the time of writing, `npm audit
+--json` reports `{prod: 1, dev: 104}` — the single prod entry being the root
+package itself — and one **HIGH** advisory (`postcss`) that the shipped
+`--omit=dev` gate never sees. All four graphs pass `npm audit --omit=dev
+--audit-level=high` today. Issue #434 tracks extending the sweep to the dev
+trees, and notes the blast radius: those dev dependencies build `farm.js` and
+`sandbox.js`, which are committed, shipped artifacts.
 
 ## Secrets scan
 
@@ -174,13 +193,27 @@ The gate is three-layered — one hosted backstop plus two cooperative local
 layers:
 
 1. `ci.yml`'s `secret-scan` job (issue #404): gitleaks' full default ruleset,
-   run over the tracked tree on every pull request as a required check for the
-   `ci-passed` merge gate. The scanner is pinned by image digest and runs in a
-   network-isolated, read-only container over a read-only mount. This is the
-   only layer a bot, a fork, a web edit, or a plain `git push` cannot skip — the
-   two below run only inside a contributor's local session. Its allowlist
-   (`.gitleaks.toml`) waives individual fake fixture literals by value, never a
-   file path: a `paths` waiver would drop the whole file from the scan.
+   run as a required check for the `ci-passed` merge gate. The scanner is pinned
+   by image digest and runs in a network-isolated, read-only container over a
+   read-only mount. This is the only layer a bot, a fork, a web edit, or a plain
+   `git push` cannot skip — the two below run only inside a contributor's local
+   session. It scans in **two modes**, because each is blind to what the other
+   catches: `gitleaks dir` over the merge-result tree (what would land on main),
+   and, on a pull request, `gitleaks git --log-opts <merge-base>..HEAD` over the
+   PR's own commits. Without the second, a credential added in one commit and
+   deleted in the next passes green and still reaches main's history under a
+   merge or rebase merge, both of which this repo allows. Findings print with
+   `--redact --verbose`, so a red job names the file, line, rule, and commit
+   while the secret itself stays `REDACTED`.
+
+   Its allowlist (`.gitleaks.toml`) waives individual fake fixture literals by
+   value, never a file path — a `paths` waiver drops the whole file from the
+   scan — and every waiver is anchored `\A<literal>\z`. The anchoring is the
+   load-bearing part: gitleaks substring-matches allowlist regexes, so an
+   unanchored literal waives every secret that merely contains it, and no choice
+   of `regexTarget` fixes that. `.github/scripts/test_ci_impact.py` enforces both
+   rules, and the `secret-scan` job re-runs that contract itself, so a path-filter
+   edit can never leave the allowlist unguarded.
 2. Manual sweep of the staged diff for credential patterns
    (`api[_-]?key|token|secret|password|private[_-]?key|passphrase|credential|BEGIN.*PRIVATE|AKIA|ghp_|sk-ant`),
    case-insensitive. This is the convenience layer; the authoritative classifier
--- a/.github/scripts/test_ci_impact.py
+++ b/.github/scripts/test_ci_impact.py
@@ -140,15 +140,83 @@ def gitleaks_allowlist_blocks(config: str) -> list[str]:
 def gitleaks_allowlist_regexes(config: str) -> list[str]:
     """Every literal inside an allowlist block's `regexes = [...]`.
 
-    Handles both the single-entry (`regexes = ['''x''']`) and multi-line array
-    spellings, so a waiver cannot hide from the narrowness contract by changing
-    its formatting.
+    Handles the single-entry (`regexes = ['''x''']`), multi-line array, and
+    embedded-newline spellings, so a waiver cannot hide from the narrowness
+    contract by changing its formatting.  The embedded-newline case is real: the
+    one multi-line secret this repo waives is a PEM block whose detected value
+    spans five source lines.
     """
     found: list[str] = []
     for block in gitleaks_allowlist_blocks(config):
         for array in re.findall(r"(?ms)^regexes\s*=\s*\[(.*?)\]", block):
-            found += re.findall(r"'''([^'\n]+)'''", array)
+            found += re.findall(r"(?s)'''(.*?)'''", array)
     return found
+
+
+# The characters that would let an anchored waiver match more than the single
+# fixed value it spells out.  `\` is in the set because an escape sequence is a
+# pattern, and this contract admits no patterns at all.
+_REGEX_METACHARACTERS = frozenset(".*+?()[]{}|^$\\")
+# `\A<body>\z` - Go RE2's spelling of "the WHOLE target is exactly <body>".
+_ANCHORED_WAIVER = re.compile(r"(?s)\A\\A(?P<body>.*)\\z\Z")
+
+
+def gitleaks_waiver_violations(config: str) -> list[str]:
+    """Every way `config`'s allowlist could waive more than one fixed value.
+
+    Empty means the allowlist is narrow.  Kept a pure function of the config
+    text so the contract can be exercised against adversarial configs that are
+    never written to disk.
+
+    Each rejected shape below was MEASURED against the pinned scanner image,
+    scanning this repo's tracked tree with a high-entropy key planted in
+    `plugins/ca/tools/.env.example` (the shipped config reports `leaks found: 1`
+    there; every shape below returns it to `no leaks found`):
+
+    * `paths` / `commits` un-scan a whole file or commit rather than waiving a
+      value, leaving a permanent blind spot.  `stopwords` is the same class.
+    * `regexTarget = "line"` widens the haystack to the entire source line, and
+      `regexTarget = "match"` to the rule's whole match text - so a short common
+      substring like `API_KEY` waives every line or match containing it.
+    * an UNANCHORED literal is the deeper bug, and the reason constraining
+      `regexTarget` alone is NOT sufficient: gitleaks tests allowlist regexes
+      with Go's `FindString`, a SUBSTRING search, so even against the default
+      (secret) target the literal `sk-` waives every secret beginning `sk-`.
+
+    Anchoring is what actually buys narrowness: `\\A<literal>\\z` matches only
+    when the ENTIRE detected secret is that exact fixture value.
+    """
+    problems: list[str] = []
+    for directive in re.findall(r"(?m)^\s*(paths|commits|stopwords)\s*=", config):
+        problems.append(
+            f"`{directive}` un-scans whole files or commits instead of waiving a value"
+        )
+    for target in re.findall(r'(?m)^\s*regexTarget\s*=\s*"([^"]*)"', config):
+        problems.append(
+            f'`regexTarget = "{target}"` binds the waiver to surrounding source text '
+            "rather than to the detected secret"
+        )
+    for block in gitleaks_allowlist_blocks(config):
+        head = (block.splitlines() or ["<empty>"])[0]
+        if "description =" not in block:
+            problems.append(f"allowlist block at {head!r} carries no rationale")
+        if "regexes = [" not in block:
+            problems.append(f"allowlist block at {head!r} names no value, so it is unbounded")
+    for literal in gitleaks_allowlist_regexes(config):
+        anchored = _ANCHORED_WAIVER.match(literal)
+        if anchored is None:
+            problems.append(
+                f"{literal!r} is not anchored `\\A...\\z`; gitleaks matches allowlist "
+                "regexes as substrings, so it waives anything merely containing it"
+            )
+            continue
+        stray = sorted(set(anchored.group("body")) & _REGEX_METACHARACTERS)
+        if stray:
+            problems.append(
+                f"{literal!r} carries regex metacharacter(s) {''.join(stray)!r}; "
+                "a waiver must spell out one fixed value"
+            )
+    return problems
 
 
 def aggregate_needs(ci: str) -> list[str]:
@@ -715,10 +783,20 @@ class WorkflowContractTest(unittest.TestCase):
         )
 
     def test_every_npm_audit_gate_fails_the_build_on_a_high_advisory(self):
-        # Issue #403 plus the maintainer fold-in: the farm and sandbox audits
-        # gated at `critical`, so every advisory cleared by #400 - all HIGH -
-        # would have sailed through.  One threshold, asserted per invocation,
-        # across every workflow that audits a dependency graph.
+        # Issue #403: site/package-lock.json - the ONLY graph in this repo that
+        # declares production dependencies (astro, starlight, markdown-remark) -
+        # was audited by nothing, and #400's three HIGH advisories all lived
+        # there.  The farm/sandbox gates were separately pinned at `critical`,
+        # but tightening those two is a DURABILITY change, not a live one:
+        # measured, plugins/ca/tools, plugins/ca-sandbox/tools and
+        # plugins/ca-pi/tools each declare ZERO production dependencies, so
+        # `--omit=dev` audits an empty graph there at any threshold.  It decides
+        # what happens the day one of them takes on a runtime dependency.
+        # Issue #434 tracks extending the sweep to the dev trees, which is where
+        # those packages' advisories actually are.
+        #
+        # One threshold, asserted per invocation, across every graph the repo
+        # ships or publishes.
         expected = "npm audit --omit=dev --audit-level=high"
         invocations: list[tuple[str, str]] = []
         for workflow in (CI_WORKFLOW, DOCS_WORKFLOW):
@@ -727,11 +805,24 @@ class WorkflowContractTest(unittest.TestCase):
                 for command in npm_audit_invocations(workflow.read_text(encoding="utf-8"))
             ]
         self.assertEqual(
-            len(invocations), 3, "expected exactly the farm, sandbox, and site audit gates"
+            len(invocations),
+            4,
+            "expected exactly the farm, sandbox, ca-pi, and site audit gates",
         )
         for name, command in invocations:
             with self.subTest(workflow=name, command=command):
                 self.assertEqual(command, f"run: {expected}")
+        # Every job that installs a lockfile must also audit it.  An unaudited
+        # `npm ci` is exactly how site/package-lock.json went unguarded, and
+        # plugins/ca-pi/tools is a PUBLISHED host package with the same shape.
+        # The Pi canary is deliberately excluded - it floats the host version on
+        # purpose and is advisory-only, so it is not a gate.
+        ci = CI_WORKFLOW.read_text(encoding="utf-8")
+        for job_id in ("tools", "ca-sandbox-tools", "ca-pi-checks"):
+            with self.subTest(job=job_id):
+                job = workflow_jobs(ci)[job_id]
+                self.assertIn("npm ci", job)
+                self.assertIn(f"run: {expected}", job, "this job installs a graph it never audits")
 
     def test_docs_workflow_gates_the_site_dependency_graph_before_publishing(self):
         # Issue #403: site-check ran npm ci/typecheck/test and build ran
@@ -784,6 +875,25 @@ class WorkflowContractTest(unittest.TestCase):
         self.assertIn("--redact", job, "a finding would print the credential in public logs")
         self.assertIn("--exit-code 1", job, "a finding would not fail the job")
         self.assertIn("--config /repo/.gitleaks.toml", job)
+        # ACTIONABLE WITHOUT LEAKING.  `--redact` alone prints only
+        # `leaks found: N` - no file, no line, no rule id - so a red job tells a
+        # maintainer nothing and the only way to triage it is to re-run the
+        # scanner unredacted.  Measured against the pinned image: adding
+        # `--verbose` prints `File:`/`Line:`/`RuleID:`/`Fingerprint:` (and
+        # `Commit:` in git mode) while still printing `Secret: REDACTED`.  Every
+        # scan step must carry BOTH, so neither mode is a dead end.
+        scan_steps = [step for step in job.split("\n      - name: ") if '"$GITLEAKS_IMAGE"' in step]
+        self.assertEqual(
+            len(scan_steps), 2, "expected exactly the tree scan and the commit-range scan"
+        )
+        for step in scan_steps:
+            with self.subTest(step=step.splitlines()[0]):
+                self.assertIn("--redact", step, "this scan would print credentials in public logs")
+                self.assertIn(
+                    "--verbose",
+                    step,
+                    "a failing scan would print only `leaks found: N` - no file, line, or rule",
+                )
         # Enforced through BOTH aggregate registrations (the #390 contract).
         self.assertIn("secret-scan", aggregate_needs(ci), "ci-passed.needs omits secret-scan")
         self.assertIn(
@@ -805,10 +915,6 @@ class WorkflowContractTest(unittest.TestCase):
         self.assertTrue(GITLEAKS_CONFIG.is_file(), "no .gitleaks.toml in the repo root")
         config = GITLEAKS_CONFIG.read_text(encoding="utf-8")
         self.assertIn("useDefault = true", config, "the config must extend the default ruleset")
-        directives = re.findall(r"(?m)^\s*(paths|commits|stopwords)\s*=", config)
-        self.assertEqual(
-            directives, [], "path/commit waivers un-scan whole files - waive by value instead"
-        )
         blocks = gitleaks_allowlist_blocks(config)
         self.assertEqual(
             len(blocks),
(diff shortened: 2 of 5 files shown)


SUaDtL:
**DO NOT MERGE** — adversarial review returned FAIL. CI is 37/37 green; that is precisely the problem.

**BLOCK 1 — the allowlist guard never runs on a diff that weakens it.** `test_the_secret_scan_allowlist_stays_narrow_and_value_scoped` lives in `.github/scripts/test_ci_impact.py`, executed only by the `ci-impact` job, which is gated behind a `changes` filter. A PR that widens the secret-scan allowlist can therefore skip the very job that guards it. The gate protects itself only by coincidence of which paths a PR happens to touch.

**BLOCK 2 — the "narrowness" contract does not enforce narrowness.** The test only requires each waiver to match `\A[A-Za-z0-9_-]+\Z`; it never constrains `regexTarget`. With `regexTarget = "line"` a fixed literal becomes a **substring** match against every scanned line — an arbitrarily broad waiver that still passes the narrowness test. Since "the waiver is narrow and value-scoped" is the entire justification for Deviation #1, the deviation is unsupported as implemented.

**HIGH — #404 AC-1 is not met, and the deviation was not declared.** The job runs `gitleaks dir .` over a `git archive HEAD` snapshot — the merge-result *tree*. It never inspects commit history, so a secret added and removed within the PR's commits is invisible. AC-1 asks for a scan over the relevant commit *range*.

**MEDIUM — a failing scan is unactionable.** `--redact` is passed without `-v` and no report is written, so the log contains only `leaks found: N` — no file, no line, no rule id. The in-workflow comment claims otherwise.

Credit where earned: the reviewer confirmed empirically that Deviation #1's central claim is TRUE (a `paths` waiver drops exactly 3,344 scanned bytes — the file's byte-for-byte size — and a planted key goes unreported), and that the AC-3 regression genuinely bites. The lane is mostly strong; these are durability defects in the new gate, not a wrong idea.

Converting to draft pending a fix round.

SUaDtL:
**DO NOT MERGE** — adversarial recheck returned FAIL. CI is 37/37 green; that is the problem again.

All four headline fixes are **real and independently proven** (pinned image resolves to gitleaks v8.30.1; planted key `sk-kR3mVq7X…` gives baseline `leaks found: 1`):

| shape | scanner | contract |
|---|---|---|
| unanchored `sk-`, `regexTarget` omitted | **swallows** | REJECTS |
| `regexTarget = "line"` | **swallows** | REJECTS |
| anchored `\A<literal>\z` | precise | ACCEPTS |
| `\Ask-` half-anchored | **swallows** | REJECTS |

So BLOCK 2's premise was right: `regexTarget` alone buys nothing — anchoring is what bites. BLOCK 1 is closed (no job-level `needs:`/`if:`, no `paths:` on `on.pull_request`, present in both `ci-passed.needs` and `required_results`). The HIGH is closed — a two-commit add-then-`git rm` gives tree scan exit 0, range scan exit 1 naming file, line, rule and commit.

## BLOCK — the narrowness contract has three bypasses

Each measured on the real tree: **guard GREEN, scanner swallows a real planted key.**

1. **Indented `regexes` key.** `gitleaks_allowlist_regexes()` extracts with `(?ms)^regexes\s*=\s*\[` — column 0 — while the "names no value" check is a substring test (`"regexes = [" in block`). Valid TOML `  regexes = ['''sk-''']` therefore extracts **zero** literals, yields zero violations, and the scanner reports `no leaks found`.
2. **`disabledRules` under `[extend]`.** `disabledRules = ["generic-api-key"]` kills the highest-value rule. The contract asserts `useDefault = true` is present but never checks what sits beside it. One idiomatic line, invisible.
3. **Single-quoted `regexTarget`.** The ban matches double quotes only (`regexTarget\s*=\s*"([^"]*)"`). `regexTarget = 'line'` is valid TOML, evades it, and paired with a whole-line-anchored literal clears the metacharacter check too.

Fix is mechanical: `^\s*regexes`, flag `^\s*regexTarget\s*=` regardless of quoting, and add `disabledRules` to the `paths|commits|stopwords` forbidden list.

## MEDIUM — `docs.yml` never runs on itself
Its triggers are `site/**` and `plugins/ca/**` only, so **the new site audit step did not execute in this PR's CI** — there is no `Site |` check among the 37. Same defect class this PR exists to fix. Verified it will not break on merge (`npm audit --omit=dev --audit-level=high` on `site/package-lock.json` → 0 vulnerabilities), but its first real run will be on an unrelated future site PR.

## LOW — `test_every_npm_audit_gate_fails_the_build_on_a_high_advisory` over-claims
Its comment says "every job that installs a lockfile must also audit it" but it iterates a hardcoded triple. `ca-pi-tools` is required, runs `npm ci`, and never audits. Benign today; a new unaudited graph would pass silently.

Verified clean: no `export-ignore` in `.gitattributes`; `gitleaks dir` does not skip gitignored files; image digest-pinned; all 13 shipped waivers anchored and metacharacter-free. Singular `[allowlist]` is already closed — gitleaks refuses it alongside `[[allowlists]]`.

SUaDtL:
Superseded by #448 — same final tree, rebuilt as a single clean commit.

**This PR's own commit-range scan caught this PR.** A real high-entropy measurement key was committed in `fc55acd` and removed in `8ec6e6d`, so the tree scan went green but the range scan stayed red on history:

```
RuleID: generic-api-key  File: .github/scripts/test_ci_impact.py  Line: 1228
Commit: fc55acd910a513c1eb14a60582614e3b9404a0f6
```

That is the control working exactly as #404 AC-1 intended — a secret added and later removed inside a PR is invisible to a tree scan, which is why the range scan exists.

Three routes existed and two were wrong: force-pushing is prohibited in this repo, and a `.gitleaksignore` fingerprint would have added a waiver surface the narrowness contract does not police — a twelfth bypass of the exact class this work closes eleven of. So the branch was rebuilt from `origin/main` as one commit with the key never entering history.

All the work is preserved in #448, including the eleven measured contract bypasses, the `docs.yml` self-trigger fix, and the derived npm-audit rule.

Issue #223, opened 2026-07-03
pre-bash.py branch guards resolve CLAUDE_PROJECT_DIR, not the worktree/cwd; heredoc prose trips command patterns


SUaDtL:
Found during the docs-site overhaul campaign (PR-0.2 author working in an isolated git worktree, then reproduced by the orchestrator).

`pre-bash.py`'s branch-sensitive guards (H-01, and the H-09b/H-14 siblings built on the same `_hooklib` primitives) resolve the repo/branch to check via `_hooklib.project_root()` / `CLAUDE_PROJECT_DIR`, which is pinned to the main checkout. Two failure modes:

1. **Worktree false positive / false negative.** A commit executed inside a linked worktree on a feature branch is judged against the MAIN checkout's branch — blocked when main is checked out there (false positive), and conversely a worktree sitting on a protected branch would be judged by the main checkout's feature branch (false negative: the protected-branch guard can be sidestepped by committing from a worktree). Workaround used in the campaign: pass the worktree path explicitly with the git `-C` flag, which the guard honors.

2. **Prose false positive.** The guard pattern-matches the whole command string, so a `gh pr create`/`gh issue create` whose heredoc BODY merely contains phrases like "git commit" or "git diff" trips H-01/H-09b even though no git write happens. Reproduced twice while opening this campaign's PRs; workaround is `--body-file`.

Fix direction: derive the repo root from the command's effective cwd (or an explicit `-C` argument) rather than `CLAUDE_PROJECT_DIR` when they disagree; and anchor the command-pattern match so quoted/heredoc payload text is not scanned as if it were the command.

https://claude.ai/code/session_01PVTCDji6bEuf9SifQ8tfsa

SUaDtL:
Second facet, hit by the PR-1.3 author: the H-01 workaround (explicit git -C <worktree>) does NOT propagate to H-09b's marker lookup — the diff scan honors -C/cwd, but the security-gate-passed marker path is still resolved from the pinned project root (CLAUDE_PROJECT_DIR), so a pass recorded by security-pass.py inside the worktree is invisible to the gate. Fix should resolve BOTH the diff root and the marker root from the same effective cwd. (The author correctly refused to hand-copy the marker into the main root; the work was landed from the main checkout instead.)

SUaDtL:
[...]

PR #458, opened 2026-07-25
fix(farm): own the child process tree and prove worktree teardown

### LINKED ISSUE #395: Verify farm timeout cleanup across the entire child process tree
# Verify farm timeout cleanup across the entire child process tree

**Severity:** high | **Confidence:** 0.95 | **Effort:** L

**Where:**
- plugins/ca/tools/exec.ts:75-89
- plugins/ca/tools/exec.ts:123-145

**Evidence:** `run()` spawns without `detached: true`. On non-Windows timeout, `treeKill` calls only `child.kill("SIGKILL")`, so grandchildren are neither addressed as a process group nor enumerated. On Windows it asynchronously starts `taskkill /T /F` but does not await its exit or verify descendants are gone; `run()` immediately resolves exit 124 after invoking `treeKill`.

**Impact:** A timed-out gate, setup command, test, or mutation hook can leave servers, watchers, and other descendants running. They can hold ports/files, mutate the worktree after the farm has moved on, leak resources across tasks, and make retries or later gates nondeterministic even though the timeout was reported as successfully killed.

**Recommendation:** Adopt the repository's verified process-tree ownership pattern for farm children: launch into an isolated POSIX process group, use a canonical absolute Windows tree-kill boundary, await bounded graceful/escalated termination, and report cleanup failure distinctly from an ordinary timeout. Do not resolve the worker timeout until descendant absence is verified.

**Acceptance criteria:**
- A live fixture that spawns a child and grandchild is timed out and all recorded PIDs are verified absent before `run()` resolves on Windows, macOS, and Linux.
- A failed or timed-out cleanup returns a distinct unhealthy/escalation result rather than ordinary exit 124.
- Repeated timeout runs leave no processes, held ports, or worktree writers behind.
- Mutation-hook and gate-command callers preserve and surface the cleanup-verification result.

### LINKED ISSUE #398: Report and retry failed farm worktree teardown before declaring green
# Report and retry failed farm worktree teardown before declaring green

**Severity:** medium | **Confidence:** 0.97 | **Effort:** M

**Where:**
- plugins/ca/tools/farm.ts:1154-1167
- plugins/ca/tools/farm.ts:1422-1424
- plugins/ca/tools/farm.ts:1900-1910

**Evidence:** Scratch sample removal and branch deletion each end in `.catch(() => {})`; a successful task similarly ignores failure from `git worktree remove --force` and immediately returns `status: "green"`. The run-level `finally` also discards integration-worktree cleanup failure. No re-list or `git worktree prune` verification follows any of these paths.

**Impact:** Locked files, Windows antivirus races, or Git failures leave worktrees and branches behind while the farm reports success. Subsequent runs then depend on best-effort destructive pre-cleanup, consume disk, retain stale source, and can fail or operate against ambiguous worktree registrations with no receipt explaining the leak.

**Recommendation:** Centralize worktree cleanup as a bounded verified operation with Windows-aware retry/backoff. Continue cleanup of all resources, retain bounded diagnostics for each failure, and make a task/run report a cleanup-degraded state (and non-zero exit when ownership cannot be released) instead of silently green.

**Acceptance criteria:**
- Injected worktree-remove and branch-delete failures are present in the task/run result and cannot produce an unqualified green exit.
- Cleanup retries bounded transient Windows lock errors and then verifies Git no longer registers the worktree and the path is absent.
- Best-of-N cleanup attempts every sample even when earlier removals fail and reports all remaining resources.
- Integration-worktree teardown failure is persisted in the final run receipt and produces a non-zero run outcome.

SUaDtL:
Closes #395
Closes #398

Lane **L01-farm-process-boundary**. Two containment defects with the same shape: a cleanup step was issued, never verified, and the run reported success one line later.

---

## Claim verification (line numbers had drifted — 19 PRs landed since filing)

| Issue claim | Verdict |
| --- | --- |
| #395 `exec.ts:75-89` / `123-145` | **REAL**, drifted to `exec.ts:79-89` (`treeKill`) and `116-147` (`run`). Every sub-claim held: spawn without `detached` (124), `child.kill("SIGKILL")` only off-Windows (84), unqualified fire-and-forget `spawn("taskkill", …)` (82), timer resolving 124 with no descendant check (136-140). |
| #398 `farm.ts:1154-1167` (sample teardown) | **REAL**, drifted to `farm.ts:1423-1426`. |
| #398 `farm.ts:1422-1424` (success path) | **REAL**, drifted to `farm.ts:1700-1702` — `.catch(() => {})` immediately followed by `status: "green"` on 1702. |
| #398 `farm.ts:1900-1910` (run `finally`) | **REAL**, drifted to `farm.ts:2414-2415`. |
| #398 "no `git worktree list`/`prune` re-verification anywhere" | **REAL** — confirmed by grep across `farm.ts`. |

**Deliberately NOT changed:** `prepareWorktree` (`farm.ts:1257-1259`) has the identical `.catch(() => {})` trio. That is stale-leftover **pre**-cleanup before a `git worktree add` that already fails loudly on its own (`worktree add failed: …` → escalate), not a teardown claim. Swallowing there is correct and is left alone.

---

## #395 — the timeout kill did not own the process tree

Every gate, setup and mutation command runs as `bash -c <cmd>` / `cmd.exe /c <cmd>`, so **the operator-authored (model-influenced) command is a grandchild**. The old kill reached only the shell.

- `run()` now spawns `detached` off Windows, so the child leads its own process group and `kill(-pid)` addresses the whole tree at once. Not on Windows, where `detached` means "new console" and breaks `cmd.exe /c`.
- `treeKill` is `async` and **verifies absence** — pid probe plus process-group probe via signal 0, polled to a budget (`FARM_KILL_VERIFY_MS`, default 10s) — before it reports success. `EPERM` reads as *present*, never as gone.
- Windows resolves `taskkill.exe` **absolutely** from `%SystemRoot%` (an unqualified name is resolved through `%PATH%` — a resolution hazard inside the containment step itself) and **awaits it to completion**. Exit 128 ("process not found") counts as success.
- A cleanup that cannot be verified is a **distinct outcome**: exit `125` + `cleanupFailed: true` + a `CLEANUP UNVERIFIED` note on both `out` and `stderr`, instead of a clean-looking 124. Both codes are non-zero, so every existing `code !== 0` consumer branch is untouched.
- `detached` is now in `RunOpts`'s `Omit` list, so the group-isolation invariant is compiler-enforced alongside the existing env-scrub and argv-array invariants.
- `run()` sets a `killing` flag *before* the (now awaited) kill — otherwise the child's own `close`, which the kill itself causes, wins the race and resolves an ordinary exit result, discarding both the `timedOut` tag and the cleanup verdict.
- `mutation.ts`'s own `spawn` gets the same treatment, and surfaces `CLEANUP UNVERIFIED` into the observability-002 "configured but failed" detail.

## #398 — teardown was best-effort and unreported

- `removeWorktreeVerified` / `deleteBranchVerified`: bounded retry with backoff (3 × 150ms, Windows lock/AV race), then re-verify against `git worktree list --porcelain` **and** the filesystem (`git branch --list` for branches — an empty listing unambiguously means absent, whereas `rev-parse`'s exit code also encodes other failures). The **verification**, not the removal command's exit code, is the authority.
- `git worktree prune` runs **only** when git still registers the target but the directory is already gone. It is repo-wide, and a blanket call would deregister an unrelated worktree whose path is merely unreachable.
- Best-of-N attempts **every** sample even after one fails — one stuck sample no longer hides the ones behind it (and a leaked sample worktree keeps its branch checked out, so the branch delete fails too; both are now reported).
- Task-owned leaks ride on the `Result` (`cleanup: CleanupOutcome[]`, present only when something failed); the run-level integration worktree rides on `RunArtifactHealth.cleanup`.
- Integration teardown now runs **before** `writeReport`, not after, so it lands in the authoritative receipt rather than a console line the operator scrolled past. Nothing in `writeReport` reads that worktree (`git diff` runs against the main checkout), so the reorder is safe.
- Following #430's tiering: releasing worktree ownership is **authoritative**, not a convenience mirror. A leak lifts the run to **exit 2** via the new pure `runExitCode()` and prints a `CLEANUP DEGRADED` block naming every unreleased resource. Exit 3 (receipt failure) still outranks it.
- The report gains a `## Resource cleanup` section and an `artifacts.cleanup.released` boolean, stated in **both** directions on purpose — silence is not evidence of a clean teardown, so an operator can tell "nothing leaked" from "this report predates the check".

`status` stays `"green"` for a task whose work genuinely merged; flipping it to `escalate` would wrongly block dependent tasks. Green now means "the work merged" and can no longer *also* mean "and we quietly abandoned a worktree" — the leak travels on the result, into the receipt, onto the summary, and into the exit code.

---

## Verbatim red (before implementation)

```
⎯⎯⎯⎯⎯⎯ Failed Tests 11 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  farm.unit.test.ts > #395 — timeout kills and VERIFIES the whole child process tree > leaves NO grandchild behind — the tree is verified gone before run() resolves
AssertionError: expected true to be false // Object.is equality
 FAIL  farm.unit.test.ts > #395 — timeout kills and VERIFIES the whole child process tree > treeKill resolves a verification result instead of firing and forgetting
TypeError: Cannot read properties of undefined (reading 'ok')
 FAIL  farm.unit.test.ts > #395 — timeout kills and VERIFIES the whole child process tree > resolves taskkill absolutely from SystemRoot (an unqualified name is a PATH hazard)
TypeError: taskkillPath is not a function
 FAIL  farm.unit.test.ts > #395 — timeout kills and VERIFIES the whole child process tree > reports an UNVERIFIED cleanup distinctly from an ordinary timeout
AssertionError: expected undefined to be true // Object.is equality
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > retries a transient remove failure and confirms git no longer registers the worktree
TypeError: removeWorktreeVerified is not a function
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > returns ok:false with diagnostics when git still registers the worktree after every attempt
TypeError: removeWorktreeVerified is not a function
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > deleteBranchVerified reports a branch git still lists
TypeError: deleteBranchVerified is not a function
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > a task whose worktree removal cannot be verified carries the failure — green is never unqualified
AssertionError: expected [] to have a length of 1 but got +0
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > best-of-N attempts EVERY sample's teardown even when an earlier one fails, and reports all leftovers
AssertionError: expected false to be true // Object.is equality
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > run exit code is non-zero when ownership could not be released, zero otherwise
TypeError: runExitCode is not a function
 FAIL  farm.unit.test.ts > #398 — verified, retried, reported worktree teardown > integration-worktree teardown failure is carried on the run health and lands in the receipt
TypeError: Cannot read properties of undefined (reading 'failures')
 Test Files  1 failed (1)
      Tests  11 failed | 182 passed (193)
```

The load-bearing one is the first: `expected true to be false` on `alive(gcPid)` **after `run()` resolved**. That is the live fixture (a hanging child that spawns a hanging grandchild, recording its pid) reproducing the actual containment defect on Windows — the grandchild was still running while the run reported a clean timeout. Every other failure is a "right reason" red: the symbol under test does not exist yet, or the field the fix introduces is absent.

## Suites — before / after

| Suite | Baseline | After |
| --- | --- | --- |
| `plugins/ca/tools` — `npx vitest run` | 287 passed, 1 skipped | **298 passed, 1 skipped** (+11 new) |
| `plugins/ca/tools` — `npx tsc --noEmit` | clean | **clean** |
| `plugins/ca/hooks/tests` | 1112 OK | **Ran 1112 tests — OK** |
| `.github/scripts` | 1110 OK, 5 skipped | **Ran 1110 tests — OK (skipped=5)** |

The whole `plugins/ca/tools` suite (including `farm.test.ts`'s real-git, real-worktree smoke tests such as *"best-of-N … cleans up scratch worktrees"* and *"is safe to run twice in a row (stale branches cleaned)"*) passes against real git with the new verified teardown in place.

`plugins/ca/tools/farm.js` — the committed esbuild artifact — is rebuilt and committed.

## NEEDS-TRIAGE

1. **Ctrl-C no longer reaches a running gate command (off Windows).** `detached: true` puts the child in its own process group, so a terminal `SIGINT` sent to farm's foreground group no longer propagates to it. Farm has no `SIGINT`/`SIGTERM` handler today, so an interrupted run now leaves the in-flight gate command running where it previously died with the terminal. This is the same leak class this PR fixes, on a path the issue did not cover; the fix is a signal handler in `main()` that tree-kills every in-flight child, which is a separate change. Windows is unaffected (`detached` is not set there).
2. **Descendant verification is a group/pid probe, not a PID census.** #395's AC says "all recorded PIDs are verified absent". We verify the process *group* is empty (POSIX) and that `taskkill /T /F` completed and the root pid is gone (Windows) — behaviourally equivalent for a tree we own, and proven by the live grandchild fixture, but it is not an enumeration of individually recorded pids. A double-fork daemon that calls `setsid()` escapes the group and would not be detected on either platform.
3. **`FARM_KILL_VERIFY_MS` is a new knob** (default 10 000ms, min 0) and is not yet documented in `.env.example` or the farm docs.
4. **`worktree list --porcelain` is re-read once per removal attempt.** On a repo with very many worktrees this is O(worktrees) per teardown. Negligible at farm's scale; noted so it is not a surprise later.
5. **Environment, not this branch:** the `.github/scripts` suite fails 4 `test_pi_benchmark` cases in a fresh worktree until `plugins/ca-pi/tools` deps are installed (`RuntimeError: Pi benchmark requires the installed pinned esbuild`). After `npm ci` there it is 1110 OK / 5 skipped. Worth a clearer skip-vs-fail decision in that test.
6. **Tooling friction (no repo change needed):** the installed plugin payload in this environment is `ca@2.8.11`, which predates #223's linked-worktree fix, so `H-01` judged the branch from `CLAUDE_PROJECT_DIR` (the main checkout, on `main`) and blocked every commit from this worktree. Worked around by addressing the repo explicitly — `git -C <worktree> commit …` — which is what `git_cwd` composes against, so the gate ran and passed on the real branch. No `--no-verify`, no override. The repo's own current hooks resolve this correctly; only the installed 2.8.11 payload is stale.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/ca/tools/exec.ts
+++ b/plugins/ca/tools/exec.ts
@@ -10,7 +10,7 @@
  * mutation.ts -> exec.ts, never back). This is a move, not a rewrite: behaviour
  * is identical to the prior in-farm.ts definitions.
  */
-import { spawn, type SpawnOptionsWithoutStdio } from "node:child_process";
+import { spawn, type ChildProcess, type SpawnOptionsWithoutStdio } from "node:child_process";
 import { readFile } from "node:fs/promises";
 import path from "node:path";
 
@@ -22,7 +22,26 @@ import path from "node:path";
 // a false `drift:` escalation. `timedOut` is set when a per-command wall-clock
 // timeout fired and the child was killed (T-06 / reliability-001) — consumers
 // surface it as a gate/setup/mutation failure rather than a clean exit.
-export type RunResult = { code: number; out: string; stdout: string; stderr: string; timedOut?: true };
+// `cleanupFailed` (#395) is set when the timeout kill fired but the child
+// process TREE could not be confirmed gone. It is a containment failure, not an
+// ordinary timeout: descendants may still be executing model-authored commands
+// against the worktree, so it carries its own exit code (see EXIT_* below) and
+// its own loud note rather than hiding inside a clean-looking 124.
+export type RunResult = {
+  code: number;
+  out: string;
+  stdout: string;
+  stderr: string;
+  timedOut?: true;
+  cleanupFailed?: true;
+};
+
+// #395: two distinguishable timeout outcomes. Both are non-zero, so every
+// existing `code !== 0` consumer branch is unchanged; the split exists so a
+// caller (and an operator reading a report) can tell "we killed it" from "we
+// could not prove we killed it".
+export const EXIT_TIMEOUT = 124; // killed, tree verified gone
+export const EXIT_TIMEOUT_UNCLEAN = 125; // killed, tree NOT verified gone
 
 // gate shell — pure determinism, no model. Use a non-login shell (`-c`, not
 // `-lc`) so user dotfiles don't bleed in. On Windows, fall back to cmd.exe /c.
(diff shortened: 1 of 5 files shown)


SUaDtL:
Cross-platform evidence for #395, since the suite table in the description is a **Windows** local run.

`[CHECK] | [CA  ] | Farm dispatcher contract` (ubuntu) is green at **299 passed (299)** — and it exercises the POSIX branch of the fix, which local Windows runs never touch (`detached` spawn + `kill(-pid)` process-group teardown):

```
✓ #395 — timeout kills and VERIFIES the whole child process tree > leaves NO grandchild behind — the tree is verified gone before run() resolves  4040ms
✓ #398 — verified, retried, reported worktree teardown > a task whose worktree removal cannot be verified carries the failure — green is never unqualified  453ms
✓ #398 — verified, retried, reported worktree teardown > best-of-N attempts EVERY sample's teardown even when an earlier one fails, and reports all leftovers  454ms
```

So the grandchild-containment claim is proven on both kill paths: taskkill/`/T /F` on Windows locally, process-group SIGKILL on Linux in CI. macOS is unexercised by this repo's matrix — the POSIX path is shared with Linux, but that is inference, not evidence (noted in NEEDS-TRIAGE).

PR #466, opened 2026-07-25
fix(governance): make an ADR's filename stem its identifier so supersession is unambiguous

### LINKED ISSUE #416: [medium] Make ADR supersession identifiers unambiguous
## Finding

The repository contains two ADR files numbered 0014:

- 0014-pi-host-authentication-and-fail-closed-tool-boundary.md
- 0014-githook-shim-dropin-fail-closed.md

Later ADR frontmatter can only say supersedes: 0014. ADR-0015 and ADR-0016 therefore have ambiguous machine-readable supersession metadata even though their prose is clear.

## Impact

/ca-adr-status and other forward-chain tooling cannot identify the intended predecessor reliably by number alone.

## Required resolution

Define a forward-compatible unique ADR identity or disambiguation rule, migrate the duplicate chain without rewriting historical decisions silently, and add parser/status tests covering duplicate legacy numbers.

SUaDtL:
Closes #416

## The defect

Two ADRs in this repository carry the number 0014, and two later ADRs supersede **different** ones of them:

| ADR | `supersedes: 0014` actually means |
| --- | --- |
| `0015-all-live-git-enforcers-and-persistent-trusted-identity` | `0014-githook-shim-dropin-fail-closed` |
| `0016-bounded-pi-child-credential-projection` | `0014-pi-host-authentication-and-fail-closed-tool-boundary` |

The same four characters name two different documents in two places, and only prose disambiguates them.

The blast radius was live, not theoretical. `_readinjectlib.py:391` and `post-write-edit.py:86` both derived the id as `fn.split("-")[0]`, so both 0014s collapsed onto one identifier. **While writing this PR my own read-inject hook emitted it:**

> `ADR-0014 (Treat Pi host authentication as opaque runtime state and fail closed on unknown tools) governs this file`

Byte-identical to what the *other* 0014 produces, apart from the title.

## The fix

**The filename stem is the identifier.** `0014-githook-shim-dropin-fail-closed`, not `0014`. It is unique by construction — the filesystem enforces it — so the collision becomes unrepresentable rather than merely unlikely.

A bare number is still accepted **only while it is unambiguous**. `0009`, `0017`, `0018`, `0019` all name one ADR and keep resolving. `0014` names two, so it raises. An ambiguous reference has no correct answer; returning one — even "the first match" — would silently invent a decision chain.

The historical collision is grandfathered by the **exact pair of stems, not by the number**. That is what makes it non-recurring: a *third* file numbered 0014 is not in the waiver and fails like any other new duplicate. The waiver covers the two files that exist, forever, and nothing else.

## TDD — verbatim red

**Red 1 — the id derivation.** Two same-numbered ADRs, indexed:

```
AssertionError: Lists differ: ['0014', '0014'] != ['0014-githook-shim-dropin-fail-closed', '0014-[49 chars]ary']

First differing element 0:
'0014'
'0014-githook-shim-dropin-fail-closed'

- ['0014', '0014']
+ ['0014-githook-shim-dropin-fail-closed',
+  '0014-pi-host-authentication-and-fail-closed-tool-boundary']
```

and the pointer text a reader would actually see:

```
AssertionError: False is not true : ['ADR-0014 (Git-hook shim) governs this file - do not contradict it; route changes via /ca:reconcile or /ca:adr.', 'ADR-0014 (Pi host auth) governs this file - do not contradict it; route changes via /ca:reconcile or /ca:adr.']
```

**Red 2 — the contract module** (16 failures, all `check_adr_identity is not importable - the ADR identity contract has no implementation yet`).

**Red 3 — the live decision record.** This one is still red on this branch; see "Blocked" below.

```
::error::ADR identity/supersession is ambiguous:
  - 0015-all-live-git-enforcers-and-persistent-trusted-identity: bare ADR number '0014' names 2 documents (0014-githook-shim-dropin-fail-closed, 0014-pi-host-authentication-and-fail-closed-tool-boundary) -- name the full filename stem instead
  - 0016-bounded-pi-child-credential-projection: bare ADR number '0014' names 2 documents (0014-githook-shim-dropin-fail-closed, 0014-pi-host-authentication-and-fail-closed-tool-boundary) -- name the full filename stem instead
```

## Proving the fix bites

- **A third file on an existing number fails.** `test_a_third_file_taking_the_grandfathered_number_is_rejected` adds `0014-a-third-decision-sneaking-in` and the contract rejects it; `test_a_new_duplicate_of_an_unrelated_number_is_rejected` does the same for `0009`.
- **An ambiguous bare number errors rather than resolving.** `test_an_ambiguous_bare_number_raises_instead_of_guessing`, plus `test_an_ambiguous_bare_number_never_returns_a_stem`, which fails loudly if a future "helpful" fallback ever picks the first match.
- **The CI path filter no longer lets it through.** The `hooks` job gains `.codearbiter/decisions/**`. Without it a PR that ONLY adds a decision file — the exact shape of a duplicate-number collision — would skip the job and land the ambiguity unchecked, the same gap that bit #384 and #404.

## The chain forks, and that is correct

`0016 -> {0017 -> 0018}` and `0016 -> 0019` are partial supersessions of different clauses. Not touched. `supersedes:` names a *document*, never a clause, so the authoring contract now says so explicitly and tells future authors to carry the scope in prose.

## Blocked: the one-time disambiguation

The maintainer's ruling, 2026-07-25:

> "the never edit rule is meant to prevent this situation, not prevent this situation from being fixed."

The principle is now recorded in `decision-lifecycle/SKILL.md`: **the never-edit rule protects decision CONTENT, not identifiers.** Rewriting what was decided corrupts the record; disambiguating which document a pointer names repairs it.

I attempted the sanctioned edit and **H-11 blocked it**:

```
PreToolUse:Edit hook error: [python3 "${CLAUDE_PLUGIN_ROOT}/hooks/pre-edit.py"]:
BLOCKED [H-11]: ADR files are edited only via /adr (ORCHESTRATOR §3) - user attribution required.
```

`.codearbiter/.markers/adr-authoring-active` is absent, and `pre-edit.py` requires it present and under 30 minutes old. I did not arm it — arming it myself would be working around the hook, not clearing it.

**To finish, arm the marker and apply exactly this:**

```
.codearbiter/decisions/0015-all-live-git-enforcers-and-persistent-trusted-identity.md  line 6
-supersedes: 0014
+supersedes: 0014-githook-shim-dropin-fail-closed

.codearbiter/decisions/0016-bounded-pi-child-credential-projection.md                  line 6
-supersedes: 0014
+supersedes: 0014-pi-host-authentication-and-fail-closed-tool-boundary
```

Simulated against a copy of `decisions/`: **exactly one line changes in each file** (line 6 both times, same line count in and out), `check_adr_identity.check()` returns `[]`, and every ADR in the chain resolves — including the 0016 fork, untouched. Not one word of any decision changes.

## Suites

| Suite | Before | After |
| --- | --- | --- |
| `plugins/ca/hooks/tests` | 1125 OK | **1125 OK** |
| `.github/scripts` | 1179 OK (5 skipped) | 1199 run, **1197 OK** (5 skipped), 2 failing |
| `tools/sync-core.py --check` | OK | **OK** |
| `tools/build-surface.py --check` | OK | **OK** |
| `tools/build-host-packages.py --check --release-guard-base` | n/a | **OK** (0.1.16 -> 0.1.17) |
| `git diff --check origin/main HEAD` | exit 0 | **exit 0** |

The 2 failures are `LiveRepositoryContractTest` — the blocked edit above, and nothing else. They go green the moment those two lines land.

## Also in this PR

- The `governs` cache gains a schema key. Its mtime stamp only invalidates on an ADR edit, so without it a payload upgrade would keep serving pre-fix numeric ids out of an already-written cache until someone happened to touch a decisions/ file.
- ca-pi 0.1.16 -> 0.1.17: the shared core moves the vendored ca-pi hooks, and a payload change on a published version never reaches installed users.
- `core/pysrc/` and `core/surface/` edited as canonical; `sync-core.py` and `build-surface.py` re-run. No vendored copy hand-edited.

### DIFF
--- a/.codearbiter/decisions/0015-all-live-git-enforcers-and-persistent-trusted-identity.md
+++ b/.codearbiter/decisions/0015-all-live-git-enforcers-and-persistent-trusted-identity.md
@@ -3,7 +3,7 @@ status: accepted
 date: 2026-07-16
 title: Require every live git enforcer and persist trusted executable identity
 decided-by: SUaDtL@users.noreply.github.com
-supersedes: 0014
+supersedes: 0014-githook-shim-dropin-fail-closed
 governs: core/pysrc/_githooks.py, core/pysrc/session-start.py, plugins/*/hooks/_githooks.py, plugins/*/hooks/session-start.py
 ---
 
--- a/.codearbiter/decisions/0016-bounded-pi-child-credential-projection.md
+++ b/.codearbiter/decisions/0016-bounded-pi-child-credential-projection.md
@@ -3,7 +3,7 @@ status: accepted
 date: 2026-07-22
 title: Permit bounded selected-provider credential projection for isolated Pi children
 decided-by: SUaDtL@users.noreply.github.com
-supersedes: 0014
+supersedes: 0014-pi-host-authentication-and-fail-closed-tool-boundary
 governs: .codearbiter/security-controls.md, .codearbiter/specs/pi-support.md, .codearbiter/plans/pi-support.md, plugins/ca-pi/**
 ---
 
--- a/.github/scripts/check_adr_identity.py
+++ b/.github/scripts/check_adr_identity.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python3
+"""Fail when an ADR identifier, or a `supersedes:` pointer, names more than one ADR (#416).
+
+An ADR is identified by its filename STEM -- `0014-githook-shim-dropin-fail-closed`,
+not `0014`. Two ADRs in this repository already carry the number 0014, and two later
+ADRs supersede different ones of them, so a bare `supersedes: 0014` names two
+documents at once and no consumer can tell which without reading prose.
+
+Invariants enforced (all derived from the repository, never hand-asserted):
+
+  1. Filename stems are unique -- the stem IS the identifier.
+  2. An ADR number resolves to exactly one stem, apart from the single
+     grandfathered historical collision below. A NEW file taking an already-used
+     number is a violation, which is what stops the collision recurring.
+  3. Every `supersedes:` value resolves to exactly one ADR. A stem always
+     resolves. A bare number resolves only while it is unambiguous; the moment it
+     names more than one stem it is an error, never a guess.
+
+Run: python .github/scripts/check_adr_identity.py   (exit 1 on any violation)
+"""
+
+import os
+import re
+import sys
+
+ADR_FILE_RE = re.compile(r"^(\d+)-.+\.md$")
+SUPERSEDES_RE = re.compile(r"^supersedes:\s*(.*?)\s*$", re.I)
+
+# Frontmatter scan depth, mirroring the hook-side indexers.
+FRONTMATTER_SCAN_LIMIT = 26
+
+# Spellings of "this ADR supersedes nothing".
+_NO_PREDECESSOR = frozenset({"", "none", "n/a", "-"})
+
+# The ONE historical number collision, grandfathered by the exact PAIR of stems
+# rather than by the number 0014. Two ADRs were authored as 0014 independently;
+# renaming either would rewrite an identifier that prose across the repository
+# already cites, and the never-edit rule protects the decision record.
+# Disambiguating the POINTERS that name them is the repair.
+#
+# Naming the stems, not the number, is what makes the collision non-recurring:
+# a THIRD file numbered 0014 is not in this set, so it is rejected exactly like
+# any other new duplicate. The waiver covers the two files that exist, forever,
+# and nothing else.
+GRANDFATHERED_DUPLICATE_STEMS = frozenset({
+    "0014-githook-shim-dropin-fail-closed",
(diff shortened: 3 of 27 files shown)


SUaDtL:
### CI state

Every job green on every OS **except** `[CHECK] | [CORE] | Hook contract`, and in that job every step passes except one — on both ubuntu-latest and macos-latest:

```
failure   Run ADR identity + supersession contract tests
skipped   Enforce the ADR identity contract on the live decision record
```

That is the blocked disambiguation, and nothing else. The two `LiveRepositoryContractTest` assertions are red because `supersedes: 0014` still names two documents; H-11 blocked the one-line-each correction and I did not arm the marker.

I moved the pair to the end of the job (dc70592) so this governance red no longer halts the run ahead of the hook-enforcement steps below it — that is why the rest of the job is now visibly green rather than merely unreached.

Arm `adr-authoring-active`, apply the two lines quoted in the description, and both go green. Verified against a copy of `decisions/`: one line changes in each file, `check()` returns `[]`.

PR #483, opened 2026-07-26
fix(farm): stop a plan secret persisting in the run's permanent receipt

### LINKED ISSUE #439: farm run-scoped receipts: allowlist plan.meta at the sink, redact artifact errors, preserve report mode
Four LOW findings from the `auth-crypto-reviewer` pass on the #430 merge. Gate verdict was **PASS** (0 CRITICAL, 0 HIGH) — none of these blocked, and #430 is already on main. Filing so they are not lost.

## 1. `plan.meta` is persisted with no control at all — and the run-scoped change widened it

`writeReport` serializes `plan.meta` verbatim (`plugins/ca/tools/farm.ts:2301`, `:2352-2359`).

Two corrections to the assumption I made when clearing this:

- **The schema is a red herring in both directions.** `plan.schema.json` is **never validated at runtime** — `farm.ts` does `JSON.parse(...) as Plan` and hand-rolls `validate()` (the file says so at `farm.ts:1710-1715`). Even `task`'s `additionalProperties: false` is enforced only by an LLM reading the schema per `writing-plans/references/farm-plan.md:39`. Adding `additionalProperties: false` to `meta` would fix **nothing**. There is currently no control — schema or runtime — on what `meta` carries into the report.
- **"Pre-existing, unchanged" is only half true.** The run-scoped change altered the exposure:
  - **Copies 1 → 2:** `plan.meta` now lands in `runs/<runId>/farm-report.json` *and* the `.farm/farm-report.json` latest pointer, plus twice more in Markdown.
  - **Retention: overwrite-per-run → permanent.** The old `.farm/farm-report.json` was clobbered by the next run. `.farm/runs/<runId>/` accumulates indefinitely with no prune path in the diff, and the surface docs fresh-id-per-run as intended usage. A secret that landed in `plan.meta` used to be destroyed by the next run; now it persists.

Stays LOW: `plan.json` is trusted operator/agent-authored input, `.farm/` is gitignored at any depth (`.gitignore:73`), farm stages only `worker.filesWritten` (the H-03 declared exception, `farm.ts:1576`), and no workflow references `.farm`.

**Fix:** allowlist-project at the sink rather than trying to close it at the schema —
`plan: { name, repo, model, apiBaseUrl, setup }` in `writeReport`. Same positive-allowlist pattern ADR-0017 mandates for the ca-pi `models.json` projection, and unlike a schema edit it is actually enforced.

## 2. New artifact error strings bypass the file's own `redactSecrets` chokepoint

`farm.ts:2244`, `:2251`, `:2255` (also `:2237`, `:2358`). Every other report-bound free-text field in this file is redacted at construction (`:302, 1303, 1309, 1320, 1422, 1446, 1520, 1608`). The new `artifacts` block is not: `git diff failed: ${d.out.trim().split("\n")[0]}` puts raw git output into `farm-report.json`, and `msgOf(e)` strings land in `artifacts.stream.errors` / `latest_mirror_errors`.

Realistically fs/git error text carrying paths and branch names, so exposure is minimal — but it is the one **new** class of report content that skips the established chokepoint. `redactSecrets` is already imported (`farm.ts:50`) and documented idempotent.

## 3. `atomicWriteFile` silently resets destination permission bits

`farm.ts:1056-1075`. The old `writeFile(dest, data)` opened an existing destination `O_TRUNC` and **preserved its mode**. The new publish creates a temp at default `0o666 & ~umask` (→ `0644`) and renames over the destination, so the temp's mode wins. An operator who hardened `.farm/farm-report.json` to `0600` gets it reset to `0644` on every run. Same effect on Windows ACLs (temp inherits the directory ACL; rename carries it).

Nothing in `security-controls.md` requires restrictive modes on farm reports — that requirement is ca-pi-scoped — hence LOW. **Fix:** `fchmod` the temp to the destination's existing mode before rename, or pass an explicit `mode`; otherwise document that report artifacts are umask-default by design.

## 4. Temp-file residue on the write/sync failure path

`farm.ts:1062-1074`. Cleanup is attached only to the **rename** failure. If `fh.writeFile` or `fh.sync()` throws (ENOSPC, EIO), the `finally` closes the handle and a partially-written `.farm-report.json.<pid>-<rand>.tmp` is left on disk holding the payload.

The unit test *"preserves the prior complete artifact and cleans up when publication fails"* only exercises the rename path (destination is a directory), so its title reads broader than its coverage.

Secondary: `open(tmp, "w")` has no `O_EXCL`, so it follows a pre-existing symlink at that path. Irrelevant under the documented single-developer model, and the random suffix makes the name unpredictable — but `"wx"` is free.

## Acceptance criteria
- AC-1: `farm-report.json` contains only the five allowlisted `meta` fields; a plan carrying an extra `meta` key does not propagate it. Pinned by a test.
- AC-2: every free-text field in the `artifacts` block passes through `redactSecrets`.
- AC-3: publishing over a `0600` report leaves it `0600`, or the umask-default behavior is documented.
- AC-4: a write/sync failure leaves no `.tmp` residue; the existing test is widened to cover that path, not just rename.

SUaDtL:
`writeReport` serialized `plan.meta` verbatim, and the run-scoped change made that **permanent** rather than transient: `.farm/farm-report.json` used to be clobbered by the next run, while `.farm/runs/<runId>/` accumulates indefinitely with no prune path. `meta.setup`, `setupEachAttempt` and `setupInputs` are raw shell-command arrays and `apiBaseUrl` is a URL — so a token in a setup command was a perfectly legal plan that landed on disk forever.

## Redaction, not an allowlist — the issue's premise was wrong

The issue states there is *"no control — schema or runtime — on what `meta` carries into the report."* There is. `checkPlanObject` is a **closed-object check**, and `plan-contract.test.ts:139` already pins that an unknown `meta` property throws at parse.

The sink allowlist the issue proposed (`{name, repo, model, apiBaseUrl, setup}`) would have been a second copy of an existing control and **caught nothing**. The exposure is the opposite shape: the *allowed* fields are the dangerous ones. So the free-text fields are redacted instead.

## What the adversarial pass broke — all reproduced end-to-end

I ran a four-lens refutation workflow against my own first cut. It found three real defects and one hollow test.

**🔴 The Markdown receipt was not redacted at all.** The first cut wrapped the JSON sink and left `farm-report.md` reading raw `plan.meta` three lines later. A run with `meta.name = "nightly sk-ant-MDPROOF-9999"` produced a redacted JSON receipt and a Markdown sibling whose H1 carried the token **verbatim**, in the same never-pruned directory, on a fully green run. The projection is now computed once and both sinks read it.

**🟠 AC-2 missed the strings it cited.** `git diff failed: <raw git stderr>`, `diffs directory unavailable: <fs error>` and `patch write failed: <msgOf(e)>` all go to `noteUnavailableDiff` — a *different* function the first cut never touched — and land in **both** receipts. Reproduced with a blocked diffs directory inside a repo path carrying a token-shaped segment.

**🟠 The redactor missed shapes a plan actually carries.** Measured misses: `glpat-`, `sk-proj-` (the `api[_-]?key` keyword doesn't fire on `OPENAI_KEY`), basic auth in a clone URL, and a bare `Bearer eyJ...`. Three are closed. The fourth is **deliberately not**: a `curl -u user:pass` rule collides with `docker run -u 1000:1000`, so the gap is recorded in the code rather than papered over.

**🟡 The mode test was vacuous on Windows.** NTFS doesn't honour POSIX mode bits, so `chmod(0o600)` reads back `0o666` and the assertion **passed against the unfixed implementation** on the maintainer's own platform. It now declares that instead of proving nothing, and keeps its teeth on POSIX CI.

## The original four ACs

`atomicWriteFile` preserves an existing destination's mode instead of resetting it to the umask default, opens its temp `O_EXCL`, and cleans up when the **write** fails rather than only on rename — the test that claimed to cover that path only exercised rename. The write failure is forced through an injectable `open` seam, because Node substitutes U+FFFD for an unencodable payload rather than throwing, so there is no "bad data" that triggers it.

## Verification

**Mutation-tested after the fixes**, since a test that can't fail is worse than none:

| mutation | result |
| --- | --- |
| identity projector | 3 failed |
| Markdown reads raw `plan.meta` | 1 failed |
| `noteUnavailableDiff` un-redacted | 1 failed |
| `noteArtifactError` un-redacted | 1 failed |

14 new tests. Full farm suite **323 passed / 1 skipped** across 5 files; typecheck clean; the shared secret-detection corpus still agrees across `redactor.ts` and `_hooklib` (`test_hooklib` 19 + 80 OK); `farm.js` rebuilt.

## Not closed, stated rather than implied

Per-task `.patch` files in the same run directory are written raw and stay that way — a redacted patch no longer applies. And redaction reduces the blast radius of a secret in a plan; it does not make `plan.meta` a safe place to put one.

Closes #439.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -153,3 +153,26 @@ high-entropy string with no trigger word, used to assert safeDiagnostic never
 fragments the audit log. Not a credential for any system.
 """
 regexes = ['''\AZx9qP2mK7vN4wR8tY1uJ6hL3sD5fA0cE\z''']
+
+[[allowlists]]
+description = """
+The three credential SHAPES the #439 redactor-reach tests assert on, in
+plugins/ca/tools/farm.unit.test.ts. An adversarial pass measured four real
+shapes reaching a permanently retained farm receipt un-redacted; three are now
+matched by the outbound redactor and one (curl's `-u user:pass`) is deliberately
+left alone because it collides with `docker run -u 1000:1000`. Each fixture
+exists to prove that boundary in a specific direction, so deleting them would
+delete the coverage:
+  sk-proj-LEAK1234567890     - must BE redacted (the api_key keyword misses
+                               OPENAI_KEY, which is why the prefix rule exists)
+  LEAKPAYLOADNOTAREALJWT123  - must BE redacted, as `Authorization: Bearer <it>`
+  ci-bot:Hunter2             - must NOT be redacted, pinning the known gap
+None is a credential for any system. Each waiver is the exact value gitleaks
+reports for that fixture, measured against the pinned image with
+--report-format json, so it can never cover anything but the literal it names.
+"""
+regexes = [
+  '''\Ask-proj-LEAK1234567890\z''',
+  '''\ALEAKPAYLOADNOTAREALJWT123\z''',
+  '''\Aci-bot:Hunter2\z''',
+]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,32 @@ predate the plugin rewrite and are grouped by date.
 
 ### Fixed
 
+- A secret in a farm plan no longer persists in the run's permanent receipt.
+  `plan.meta` was serialized verbatim into `.farm/runs/<runId>/farm-report.json`
+  and its Markdown sibling, and the run-scoped change made that permanent: the
+  old `.farm/farm-report.json` was clobbered by the next run, while
+  `.farm/runs/` accumulates with no prune path. `meta.setup`,
+  `setupEachAttempt` and `setupInputs` are raw shell-command arrays and
+  `apiBaseUrl` is a URL, so a token in a setup command was a perfectly legal
+  plan that landed on disk forever. Every free-text `meta` field is now redacted
+  at the sink, once, for both receipts. Note this is redaction and not an
+  allowlist: `checkPlanObject` is already a closed-object check that rejects
+  unknown `meta` keys at parse, so the allowlist the issue proposed would have
+  duplicated an existing control and caught nothing — the *allowed* fields were
+  the exposed ones (issue #439).
+- The artifacts block and the diff-evidence reasons now pass through the file's
+  own redaction chokepoint, so raw `git` stderr and filesystem error text cannot
+  reach either receipt un-redacted.
+- The outbound redactor recognises four credential shapes an adversarial run
+  proved were reaching a receipt intact: GitLab PATs (`glpat-`), OpenAI project
+  keys (`sk-proj-`), basic auth embedded in a clone URL, and bearer tokens. A
+  `curl -u user:pass` rule is deliberately *not* added — that shape collides
+  with `docker run -u 1000:1000`, and the gap is recorded rather than papered
+  over.
+- `atomicWriteFile` preserves an existing destination's permission bits instead
+  of resetting them to the umask default, opens its temp file `O_EXCL`, and
+  cleans up when the *write* fails rather than only when the rename does. The
+  test that claimed to cover that path only exercised the rename.
 - The shipped `farm.js` bundle is now executed by the test suite, not merely
   regenerated and byte-compared. `includes/farm.md` tells operators to run
   `node <plugin>/tools/farm.js`, but the integration launcher ran `farm.ts`
--- a/plugins/ca/tools/farm.js
+++ b/plugins/ca/tools/farm.js
@@ -188,7 +188,33 @@ async function readWorktreeFile(wt, relPath) {
 }
 
 // redactor.ts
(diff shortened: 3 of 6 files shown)


SUaDtL:
Superseded by #487, which is the identical change as a single clean commit.

The tree scan was green, but the **commit-range** scan on this branch was not: the first cut committed a realistic JWT-shaped fixture (`eyJ...LEAKPAYLOAD.sig`), and that scan correctly reports a secret introduced by a commit even when a later commit removes it. It cannot be waived narrowly either — a JWT contains dots, and a dot is a banned metacharacter in an anchored `.gitleaks.toml` waiver body.

Force-pushing to rewrite the history is prohibited here, so the change was rebuilt on a fresh branch instead. That is the honest fix: the literal should never have been committed in that shape.

Issue #511, opened 2026-07-26
plugins/ca/tools is below the stage-2 coverage floor on lines and branches


SUaDtL:
Split out of #507 so that issue can close on making the coverage gate *runnable*, rather than expanding into "write several hundred statements of tests before anything ships."

## Measured state

Taken with `@vitest/coverage-v8@4.1.9`, ephemeral install (`--no-save`), nothing adopted:

| tree | Lines | Branches | Statements | Functions |
| --- | --- | --- | --- | --- |
| `plugins/ca/tools` | **67.22%** (798/1187) | **59.46%** (575/967) | 63.91% (889/1391) | 56.77% (134/236) |
| `plugins/ca-pi/tools` | 85.37% | 78.73% | 80.12% | 85.31% |
| `plugins/ca-sandbox/tools` | 86.13% | 79.96% | 85.49% | 81.59% |

The threshold is **lines and branches, both binding** at the `stage:` value in `CONTEXT.md` (currently 2 → 70%), per `plugins/ca/includes/maturity-coverage.md`.

`ca/tools` is the only tree below it, and it is below on both binding columns. ca-pi and ca-sandbox clear 70 on both today.

## Why it matters here specifically

`plugins/ca/tools` holds `farm.ts` (152 KB) — the farm dispatcher that `/ca:sprint --farm` runs work through — plus `exec.ts`, `mutation.ts`, `redactor.ts`, and `worktree-fs.ts`. It has 324 tests, so this is not an untested tree; it is a tree whose largest module has thin branch coverage.

Branch coverage at 59.46% means roughly two in five conditionals have one side never taken. #504 is the worked example of what that hides: a `catch` block with **zero** executions inside a 750-test suite, which line coverage would have flagged and no assertion did.

## Acceptance criteria

- AC-1: `plugins/ca/tools` reaches **≥ 70% lines and ≥ 70% branches**, measured by the coverage command named in `.codearbiter/tech-stack.md`.
- AC-2: the backfill is driven by the uncovered report, not by chasing the number — each added test pins a real behaviour, and no test is added solely to move a percentage. A test that asserts nothing meaningful is worse than the gap it closes.
- AC-3: error and refusal paths are prioritised over happy paths. Those are the ones a green suite hides, and the ones `farm.ts` has most of.
- AC-4: no pre-existing test is modified to raise coverage.

## Suggested approach

[...]

Issue #504, opened 2026-07-26
flaky: live Git Bash background job resolves with no handle on windows-latest


SUaDtL:
## The flake

`test/background-jobs.test.ts > session-local background job state > launches and disposes a real Git Bash background job` fails intermittently on `windows-latest`:

```
AssertionError: expected undefined to be defined
  at background-jobs.test.ts:603
```

Observed on PR #503, Pi 0.80.10 / windows-latest. A re-run of the identical commit passed, so it is intermittent rather than a regression. The PR that surfaced it changed **only** Markdown, CHANGELOGs and version strings — no executable code — so nothing in the diff can reach a test that spawns a real Git Bash process.

## Why this is NOT the same bug as #495

Worth separating, because the obvious reading is wrong and would lead to the wrong fix.

#495 was a **deadline** problem: a 5s vitest timeout could not cover a cold CPython launch, and the symptom was `Test timed out in 5000ms`.

This one is not a timeout at all. The assertion that fails is:

```ts
const job = await awaitLiveCondition(
  runtime.launch({ ... }),
  "the live Git Bash launch",
  () => "no job handle - the contained launch never returned",
);
expect(job).toBeDefined();          // <- line 603, this is what fails
```

`awaitLiveCondition` **rejects** with a descriptive `did not settle within Nms` error when it times out. That error was not raised. So the helper *resolved* — with `undefined` — comfortably inside its 60s `WINDOWS_LIVE_TEST_TIMEOUT_MS` budget.

In other words: `runtime.launch(...)` returned **no job handle**, promptly and without erroring. Raising the timeout would change nothing.

## What that points at

A launch that resolves to `undefined` rather than throwing is a spawn that did not produce a tracked job. Candidates worth checking, in rough order:

1. **`shellPath` resolution.** The test resolves a real Git Bash. If that lookup returns a path that is momentarily unusable on the runner image, a launch could fail soft.
2. **`createBackgroundJobRuntime(...)` returning a runtime whose `launch` can resolve empty** — note the `!` on its construction; if the failure is upstream of `launch`, the handle is simply never registered.
3. **A race between process start and job-table registration** — the handle is read before the tracker has recorded it, so the read is legitimately empty rather than late.

(3) would be the most concerning of the three, because it means the runtime can report success while owning nothing — and this is the module that tracks background jobs for cleanup, so an untracked job is a leaked process.

## Why it matters beyond a re-run

[...]

PR #513, opened 2026-07-26
fix(ca): the coverage gate could not run, in any tree (#507)

### LINKED ISSUE #507: tdd Phase 5 cannot run: no coverage command exists for any plugin
## The gap

`tdd` Phase 5 is a BLOCK gate that runs "the coverage command from `tech-stack.md`" and compares the result against the maturity threshold (stage 2 → ≥ 70%, per `plugins/ca/includes/maturity-coverage.md`).

For `plugins/ca-pi/tools` that command does not exist:

- `.codearbiter/tech-stack.md` has no `## Coverage` section and names no coverage invocation for any of the four plugins.
- `plugins/ca-pi/tools/package.json` has three scripts — `build`, `typecheck`, `test` — and no coverage script.
- No coverage provider is installed: `node_modules/@vitest/` contains `expect`, `mocker`, `pretty-format`, `runner`, `snapshot`, `spy`, `utils` — no `coverage-v8`, no `coverage-istanbul`.
- `plugins/ca-pi/tools/vitest.config.ts` configures no `coverage` block.

The `tdd` hard rule is explicit: *"MUST NOT guess the test, coverage, or lint command — read `tech-stack.md` or STOP."* So every ca-pi change reaches Phase 5, finds nothing to run, and passes through on a gap rather than on a measurement.

## Why it matters

This is not a missing number, it is a **silently non-functional gate**. Phase 5 reads as satisfied in every ca-pi lane, which is the same failure shape as the release lanes in #501/#502/#506: a gate that appears green because it never actually ran for its caller's arguments.

The threshold table exists and is referenced by two skills (`tdd` Phase 5, `refactor` Phase 2). Neither can enforce it here.

Surfaced while fixing #504, where the gap blocked Phase 5 rather than being papered over.

## Scope

Same question applies to the other three plugins' TypeScript trees — `plugins/ca/tools`, `plugins/ca-sandbox/tools` — and to the Python hook suites, which run under `unittest` with no coverage measurement either. Worth deciding once across all of them rather than per-plugin.

## Acceptance criteria

- AC-1: a decision on whether coverage is measured at all in this repo. "We deliberately do not measure coverage" is a legitimate answer — but then `tdd` Phase 5 and `refactor` Phase 2 must say so instead of pointing at a command that does not exist.
- AC-2: if measured — a coverage provider is added via `/ca:add-dep` (license/provenance/supply-chain vetting; it is a new dev dependency in a plugin that ships a bundled artifact), a `coverage` script exists, and `tech-stack.md` names the exact invocation per plugin.
- AC-3: if measured — the stage-2 threshold (≥ 70%) is either met or the shortfall is recorded as a decision in `CONTEXT.md`, per the `tdd` hard rule forbidding an unrecorded threshold change.
- AC-4: if not measured — `maturity-coverage.md`, `tdd` Phase 5, and `refactor` Phase 2 are reconciled so no skill claims to run a gate that cannot run.

## Relevant sources

- `plugins/ca/skills/tdd/SKILL.md` — Phase 5, and the hard rule against guessing the command
- `plugins/ca/includes/maturity-coverage.md` — the threshold table
- `.codearbiter/tech-stack.md` — `## Test` and `## Lint / typecheck` present, no coverage section
- `plugins/ca-pi/tools/package.json`, `plugins/ca-pi/tools/vitest.config.ts`

SUaDtL:
Closes #507. Step 3 of the chain; **stacked on #510** (base is `chore/vitest-4-ca-tools`, so the diff stays clean — GitHub retargets to `main` when #510 merges).

## The defect

`tdd` Phase 5 and `refactor` Phase 2/6 instruct *"run the coverage command from `tech-stack.md`"*, and both carry a hard rule forbidding guessing one. No coverage command existed — no provider in any of the three TypeScript trees, no script, and `tech-stack.md` contained the word "coverage" **zero times**.

So every run reached the phase, found nothing to run, and passed through on a gap. **A BLOCK gate that cannot execute is worse than an absent one** — it reads as satisfied in every lane. Same defect class as #501 (five suites running on nothing) and #506 (an assertion that restated the bug).

`refactor` was the worse casualty. Phase 2 is *"Behavioral parity coverage proof"* — the gate that justifies the entire lane by proving pre-existing tests can detect a behaviour change **before** production code is touched. It rested wholly on a command that did not exist, leaving the lane running on Phase 5's "zero pre-existing tests modified" alone.

## What lands

**Tooling** — `@vitest/coverage-v8@4.1.9` in all three trees, `npm run coverage` in each, scope/provider/reporters in each `vitest.config.ts` so the invocation is argument-free and platform-identical. `ca/tools` gets its first `vitest.config.ts`; coverage output is gitignored.

**The metric, named.** The threshold table said only "minimum coverage". That omission was load-bearing, not cosmetic — a report gives four numbers that disagree:

| tree | lines | branches | statements | functions |
| --- | --- | --- | --- | --- |
| `plugins/ca/tools` | **67.22%** | **59.46%** | 63.91% | 56.77% |
| `plugins/ca-pi/tools` | 85.37% | 78.73% | 80.12% | 85.31% |
| `plugins/ca-sandbox/tools` | 86.13% | 79.96% | 85.49% | 81.59% |

ca-pi is compliant at maturity 3 on lines and non-compliant on branches. "≥ 85%" with no column named is not a threshold anyone can be held to.

Now **lines and branches, both binding**. Lines catches code no test reaches — which is exactly what #504 turned out to be, a `catch` with zero executions inside a 750-test suite. Branches catches the untaken half of a condition a test *does* reach, which lines alone reports as covered. Statements duplicates lines; functions is noisiest at small counts.

**A guard against recurrence.** `GateCommandTest` asserts that any command a gate reads from `tech-stack.md` is actually defined there. Verified in both directions rather than assumed — run against the pre-fix file it **fires**, and it correctly identifies both demanders (`tdd`, `refactor`).

## Deliberate non-choices

- **Threshold not encoded in the vitest configs.** The skills apply it from `CONTEXT.md` `stage:` against `maturity-coverage.md`. Copying the number into three package configs forks the source of truth, and the copies drift the first time the stage moves.
- **No CI job enforces coverage.** `ca/tools` is below the stage-2 floor, so wiring it into required CI would block every merge in the repo on an unrelated backfill. It is an orchestrator gate, which is what the skills already assume.
- **`stage:` stays at 2.** Raising to 3 today blocks all three trees — ca-pi fails branches at 78.73, ca-sandbox at 79.96. A gate nothing can pass gets overridden, and an overridden gate teaches readers it is noise.
- **The ca/tools shortfall is #511, not this PR.** Folding it in would turn #507 into "write several hundred statements of tests before anything ships".
- **Python hooks stay unmeasured.** Both skills now say so explicitly, so the phase is recorded as having no numeric floor rather than silently skipped.

## Versioning

The surface renders into **three hosts**, so this is a payload change for ca, ca-codex and ca-pi — not just a `tools/` edit.

- **ca-pi 0.1.34 → 0.1.35** + changelog + regenerated root manifest. Its guard demands a strict advance on any payload change regardless of tag: `Pi payload, version, changelog, and root metadata advanced together: 0.1.34 -> 0.1.35`.
- **ca (2.9.1) and ca-codex (0.3.0)** are unpublished (no `v2.9.1` / `ca-codex-v0.3.0` tag), and their gates permit a payload change on an unpublished version. Entry added under root `## [Unreleased]`.

## Verification

- `npm run coverage` green in all three trees; ca/tools still discovers all 324 tests after gaining a config
- 23 python suites + `unittest discover` over `plugins/ca/hooks/tests` + `check-plugin-refs.py`
- `sync-core.py --check`, `build-surface.py --check`, `build-host-packages.py --check` — all in sync
- `test_pi_package.py`'s exact workspace contract updated for the new script and dependency
- `tsc --noEmit` clean in all three trees

`build-surface --check` caught me editing rendered files instead of `core/surface/` templates; the edits were moved to source and re-rendered.

https://claude.ai/code/session_01JgdLb6n8mUdkFiKDTA37ML


SUaDtL:
Replaces #512, which GitHub auto-closed when `--delete-branch` on #510 removed its base branch. A closed PR cannot be retargeted or reopened, so this is the same commit (`2681e57`) plus a merge of `main`.

**Merge resolution, for the record:**

- `.codearbiter/gate-events.log` — resolved as an **append-only union** (1009 rows from main, 16 branch-only rows appended). `--ours` and `--theirs` each discard real audit rows.
- `plugins/ca/tools/package-lock.json` — took this branch's superset (vitest 4 + coverage-v8), then regenerated with `npm install --package-lock-only` so it is npm's own output rather than a hand-merged lock. Verified: declared and locked versions agree at 4.1.9 for both packages.

Re-verified after the merge, not just before it:
- clean `npm ci --ignore-scripts` + `npm run build` → `farm.js` **unchanged**
- `npm run coverage` in ca/tools → 324 tests, 67.22% lines / 59.46% branches (unchanged, as expected)
- 21 python suites + hooks unittest + all four generator `--check`s → pass
- ca-pi release guard against the **new** main: `0.1.34 -> 0.1.35`

#509's `--ignore-scripts` guard and this PR's `GateCommandTest` now coexist in `test_ci_impact.py`; the suite passes with both.

Issue #515, opened 2026-07-26
flaky: a task escalates under full-suite load, flipping exit 0 to 2 in the farm artifact-publication tests


SUaDtL:
## The flake

`plugins/ca/tools/farm.test.ts > farm artifact publication (#397 / #387) > #387: the unavailable-diff list is bounded and still reports the true total` fails intermittently under **full-suite** load:

```
AssertionError: expected 2 to be +0 // Object.is equality
 ❯ farm.test.ts:1694:37
   expect(result.code, result.out).toBe(0);
```

The captured `result.out` on the failing run:

```
Done. green=11 escalate=1 blocked=0
```

Observed 2026-07-26 on `main` @ `c9c5cbc`, `windows`, during a `npm --prefix plugins/ca/tools run coverage` run. Re-running the same test **in isolation** passed, and a second full-suite run passed. So it is intermittent and load-sensitive rather than a regression.

## Why the exit code moved

`runExitCode` (farm.ts) maps `escalated || blocked || aborted || cleanupFailures` to 2. The test plans **12 tasks at `FARM_CONCURRENCY: "6"`** against the in-process mock server. On the failing run, 11 tasks went green and **one escalated**, so exit 2 is a truthful report of what happened — the flake is that a task escalated at all, not that the exit mapping is wrong.

That points at contention rather than at the assertion: `farm.test.ts` runs alongside four other suites, several of which also spawn subprocesses, and this is the only case in the file that drives six concurrent workers.

## Why it is not #504

#504 is a `ca-pi` `background-jobs.test.ts` failure — a live Git Bash launch resolving with no handle on `windows-latest`. Different tree, different module, different symptom. Filed separately so neither investigation inherits the other's wrong lead.

## Why it matters

It reddens a suite that merge decisions are gated on. A green CI is one of the two standing conditions for merge authority in this repo, and a test that fails once per few full runs trains readers to re-run rather than read — the habit the gate exists to prevent. It cost one full coverage cycle while establishing the #511 baseline.

## Acceptance criteria

- AC-1: the escalation is explained — specifically whether a task escalates under six-way concurrency because of mock-server contention, a gate timeout under load, or a genuine race in the dispatcher's concurrency handling.
- AC-2: if the cause is test-harness contention, the test is made deterministic without weakening what it asserts (it must still prove the unavailable-diff list is bounded at 11 while reporting a true total of 12).
[...]

PR #516, opened 2026-07-26
test(farm): pin the redactor's PEM span boundaries in-process (#511)

SUaDtL:
Slice 1 of 6 on #511. Plan approved and committed here as `.codearbiter/plans/coverage-backfill-ca-tools.md`.

## What this pins

`farm.test.ts:799` already proves PEM span redaction end-to-end. But it drives the dispatcher through `spawn()`, and **the v8 coverage provider does not instrument child processes** — so the span logic read as *zero executions* in the report while being genuinely tested. The gap was real, just not the one the report described: the span's **boundaries** had no coverage from anywhere.

Boundaries are where a span redactor fails quietly. Too narrow and key body lines ship. Too greedy and it eats the rest of the file while still looking like it worked — the key *is* redacted, so a test asserting only "no key body in the output" passes either way.

Added cases, in both directions:

- content after `-----END-----` survives, and the span stops exactly there
- an unterminated block redacts through end-of-content (truncated armor is still key material)
- two blocks stay separate, with the line between them intact
- `-----BEGIN CERTIFICATE-----` — no trigger word anywhere — still spans
- an inline delimiter mention falls through to per-line handling instead of opening a span
- `isSecretBearingFilename` is a **basename** rule: `.env/notes.md` and `src/id_rsa/index.ts` must pass, which a substring implementation would deny

## Mutation evidence

Per the plan's test standard, every assertion was verified to **fail against a mutated source**, not merely to pass. Control run on unmutated source green first — a harness whose control fails proves nothing.

```
control: PASS (unmutated suite green)

negate PEM_BEGIN guard (span never entered)             KILLED
drop PEM_END test (span runs to end-of-content)         KILLED
off-by-one: span consumes the line AFTER end            KILLED
remove `continue` (END line falls through to per-line)  KILLED
unanchor PEM_BEGIN (inline mention starts a span)       KILLED
basename -> first segment in denylist match             KILLED
denylist .some -> .every                                KILLED
drop case-insensitive flag on the .env rule             KILLED

8/8 killed
```

## Coverage

`redactor.ts`: **77.77% → 100% lines, 50% → 87.5% branches**, 100% statements.

The single residual branch is `redactor.ts:106`'s `?? relPath`, which is **unreachable**: `String.prototype.split` never returns an empty array, so `.pop()` is never `undefined`. It exists only because TypeScript types `pop()` as `T | undefined`. Not chased, and recorded in the plan as known-unreachable.

Tree: 67.22 → 67.56 lines, 59.46 → 59.77 branches. Small by design — slice 1 is deliberately cheap so the file convention and the mutation bar are validated before slices 3–5 scale them.

## Verification

- `npm --prefix plugins/ca/tools run coverage` — 344 passed, 1 skipped, 6 files
- `npm run typecheck` — clean
- `npm run build` + `git diff --quiet -- farm.js` — clean, no stale bundle
- 15 Python suites from `tech-stack.md` + `unittest discover -s plugins/ca/hooks/tests` — all pass
- `check-plugin-refs.py` — intact

No version bump: `payload_scope.py` excludes `plugins/ca/tools/` from `ca`'s payload except `farm.js`, which is unchanged.

**#511 AC-4 holds — no pre-existing test modified.** This adds one new file.

## Note on CI

#515 (filed today, amended) is a pre-existing flake in `farm.test.ts`'s artifact-publication block that reddened two of three full-suite runs during this work — a task escalates under load and flips exit 0 to 2. It is independent of this change (first observed on clean `main` @ c9c5cbc before any file existed here). If CI is red on that test, it is #515, not this PR.

Refs #511


SUaDtL:
Superseded by #517.

The first commit here (`9b8025f`) carried a literal PEM header, and while the fix in `2ce17f6` cleared the **tree** scan, the **range** scan correctly kept reporting the literal at that commit — the documented two-scan behaviour in `ci.yml`, which exists because this repo allows merge and rebase merges, so a PR's individual commits reach `main` verbatim.

No follow-up commit can clear that; only history can, and force-push is barred by H-02. #517 is the same work rebuilt from `main` as a single clean commit, with the adversarial review's findings folded in rather than stacked on top.

Issue #521, opened 2026-07-27
coverage figures are host-dependent, but tech-stack.md says the command is 'identical on every platform'


SUaDtL:
## The claim

`.codearbiter/tech-stack.md`, `## Coverage`:

> Each prints a text summary and writes an html report to that tree's `coverage/` … the script takes no arguments so **it is identical on every platform**.

True of the *command*. Misleading about the *result*, which is what a reader is actually deciding with — and `tdd` Phase 5 gates on that result.

## Measured divergence

Two independent cases found while working #511, both in `plugins/ca/tools`:

**1. `worktree-fs.ts` — an over-long path segment reaches a different guard on each platform.**
`writeWorktreeFile(root, "d".repeat(300) + "/x.txt", …)` on Windows fails at `mkdir` with a non-EEXIST error and takes the `throw mkdirError` arm. On Linux `lstat` fails first with `ENAMETOOLONG` and takes the `!isMissing(error)` arm instead. Same input, different branch, one of them uncovered on each host.

Net: **83.33% branches on Windows vs 84.61% on Linux** for that file alone.

**2. `mutation.ts` — a hook fixture that only parses on one platform.**
A test setting `FARM_MUTATION_CMD` to `echo {"score":0.75,"total":8}` passes on Windows and fails on Linux: the hook runs under `bash -c`, which **brace-expands** an unquoted `{…}` containing commas, so the score line never parses. `cmd.exe` does not brace-expand. Green locally, red on CI. (Fixed in #520 by single-quoting; the divergence is the point here, not the fixture.)

## Why it matters now

**CI runs the `tools` job on `ubuntu-latest` only** (`.github/workflows/ci.yml`). The coverage gate, by contrast, is an *orchestrator* gate — `tdd` Phase 5 runs it wherever the developer is. So the number that gates a change and the number CI would produce are measured on different platforms, and this issue documents that they do not agree.

#511's AC-1 says the tree must reach "≥ 70% lines and ≥ 70% branches, measured by the coverage command named in `.codearbiter/tech-stack.md`". It does not name a platform. At the current margin that is not academic: the two hosts differ by more than a point on a single file, and #511 is being driven to a threshold from below.

## What would fix it

Options, roughly in increasing cost:

1. **Say so.** Replace "identical on every platform" with a note that the *command* is identical while the *report* is host-dependent, and name which host a recorded figure was taken on. Cheapest, and enough to stop the wrong inference.
[...]

PR #524, opened 2026-07-27
test(ca): cover farm.ts's exported surface to clear the stage-2 branch floor (#511)

SUaDtL:
Slice 5 of #511. `farm.ts` was the last module below the stage-2 floor.

Almost all of its uncovered arms sit behind `main()`, which is reachable only through `spawn()` — a subprocess the v8 coverage provider does not instrument. So this drives the **exported surface** instead: `runTask` through its `RunTaskDeps` seam, `buildPrompt`, `validate`, `parsePlan`, and — via `runTask` against a real worktree — the enrichment and byte-cap path that decides what crosses the trust boundary.

**Windows:** branches 66.18% → 72.59%, lines 72.03% → 75.82%. The authoritative Linux number is posted in a follow-up comment; Linux runs about a point below Windows on this tree, which is issue #521.

### What the tests assert

Every escalation asserts its **exact note string**, never just `status === "escalate"`. `runTask` has eleven distinct escalation paths, and a status-only assertion lets any two of them collapse into each other — a mutant that swaps one note for another survives. Thresholds and truncation widths are pinned to their numbers (`.slice(0, 200)`, the `evaluated >= 5` floor, the strict `< warnBelow` boundary), not their shapes.

### Mutation testing

46 single-point mutants plus 3 property-level ones. **Four real gaps were found and fixed this way**, three of them the same defect: an assertion that could not tell a layered guard from the layer behind it. The sample-level containment sweep and the sample tamper guard both *looked* covered, because disabling them produced an identical note from the task-level layer. They now assert an outcome only the inner layer can produce — a clean sibling sample winning, rather than the whole task escalating.

Remaining survivors are classified rather than ignored:

| survivor | classification |
|---|---|
| `t.model ?? model`, cleanup attach, `noteCleanup`, prepErr attempts | killed by the pre-existing suites — verified, not assumed |
| `Math.max(1, …)` on `FARM_SAMPLES` | provably dead: `numEnv(…, {min: 1})` returns ≥ 1 on every path |
| the two byte-cap boundary arms | need the injected total to land within ~90 bytes of the cap; not reachable without replicating the production byte arithmetic in the test |

Closes part of #511.

### DIFF
--- a/.codearbiter/plans/coverage-backfill-ca-tools.md
+++ b/.codearbiter/plans/coverage-backfill-ca-tools.md
@@ -170,6 +170,51 @@ a factor of three. `mutation.ts` was sized against its uncovered report before c
 a number this time: its arms are business logic, not layered guards, and `MUT` is an exported
 mutable object, so its knobs are settable from a test with no module surgery.
 
+## Slice 5 — `farm.ts` exported surface: CLEARED, on both platforms
+
+| | before slice 5 (Win) | after (Win) | before (Linux) | **after (Linux)** | floor |
+|---|---|---|---|---|---|
+| Lines | 72.03% | 75.82% | 70.93% | **74.72%** | 70% |
+| Branches | 66.18% | 72.59% | 65.35% | **71.76%** | 70% |
+
+**AC-1 is met on both platforms**, which dissolves #521 as a *dependency* of #511: whichever
+platform is declared authoritative, the answer is the same. #521 remains open on its own merits —
+the divergence is real (`exec.ts`: 87.50% Win vs 76.38% Linux) and will bind again at the next floor.
+
+The slice-5 estimate of "~68 available, need 38" was **wrong in both terms, and wrongly framed as
+tight**. It counted only `runTask` (36) + `validate` (12) + small helpers, and missed the nested
+closures — `bestOfN`'s per-sample body (8 arms) and `validate`'s `checkSetupInputs` (8) — which the
+per-function attribution only surfaces if you resolve each branch to its *innermost* enclosing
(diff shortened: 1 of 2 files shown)


SUaDtL:
## Authoritative Linux measurement

Clean `git archive` export of this commit + fresh `npm ci` in `node:24` — no dev-tree state, no cached `node_modules`.

|  | Windows | **Linux (CI platform)** | floor |
|---|---|---|---|
| Lines | 75.82% | **74.72%** | 70% ✅ |
| Branches | 72.59% | **71.76%** | 70% ✅ |

482 tests pass, 1 skipped, on Linux.

### This resolves the #521 blocker for #511

#521 asked which platform is authoritative for AC-1, because the two measurably disagree and AC-1 names a coverage command but no platform. That question mattered while the tree was clearing on one platform and not the other.

It no longer gates #511: **both platforms now clear both floors**, with roughly 1.8 points of margin on the tighter one. Whichever platform is declared authoritative, the answer for #511 is the same.

#521 is still worth settling on its own merits — the divergence is real and will matter again at the next floor — but it is no longer a dependency of this issue.

### Where the remaining Linux gap sits

`farm.ts` is at 68.25% branches and is the only module still below the floor individually; the tree clears because the other four carry it. The rest of `farm.ts` is `main` (67 arms), `writeReport` (18), `runCanary` (15), and `callApi` (14) — all private and reachable only through `spawn()`, which the v8 provider does not instrument. Raising those needs a harness change (running the dispatcher in-process), not more tests, and is out of scope here.

One divergence worth noting for #521: `exec.ts` reports 87.50% branches on Windows and 76.38% on Linux. That is the same platform-specific containment code (`taskkill` vs POSIX signals) measured from opposite sides, not a regression.

PR #526, opened 2026-07-27
fix(farm): make an unreported mutation count representable, not invented (#525)

### LINKED ISSUE #525: farm.ts's mutation-escalation note reports `evaluated` under a "survived" label
`farm.ts:1955`, on the hard-escalation arm of the mutation guard:

```ts
riskNote = `gaming: mutation score ${mut.score.toFixed(2)} (${mut.evaluated} mutants survived — the test does not constrain the implementation)`;
```

`mut.evaluated` is the number of mutants **evaluated**, not the number that **survived**. The two are equal only when the score is exactly 0.

The warn arm one branch down gets it right:

```ts
riskNote = `mutation-risk: score ${mut.score.toFixed(2)} (${mut.survivors.length}/${mut.evaluated} survived) — weak test or under-implemented logic`;
```

### Why it matters

This note is the escalation reason an operator reads in the farm report — it is the whole explanation for why a task was rejected. With `{evaluated: 10, killed: 1}` it currently reads "mutation score 0.10 (10 mutants survived)" when 9 survived. It overstates the failure, and it contradicts the score printed in the same sentence: a reader who does the arithmetic gets a different number than the one they are shown, which undermines trust in a signal whose entire job is to be believed.

### Suggested fix

Mirror the warn arm: `${mut.survivors.length}/${mut.evaluated} survived`. That also makes the two arms consistent with each other, which they are not today.

### Found by

Adversarial review of #524 (slice 5 of #511). Not fixed there: that PR is test-only, and touching `farm.ts` forces a `farm.js` bundle rebuild, which **is** declared payload and would pull in a `ca` version bump plus a `farm-dist.test.ts` parity re-check.

`plugins/ca/tools/farm-surface.test.ts` currently pins the **wrong** string deliberately, with a comment pointing here, so that current behaviour is documented rather than silently accepted. Fixing this line will turn that test red — that is intended; update the expected string in the same change.

SUaDtL:
Root-cause fix for #525, reached via `/ca:debug` after three symptom fixes each produced a *new* wrong count. Five adversarial review rounds; this body describes `8ce64b6`.

### The defect and why it kept moving

`farm.ts`'s escalation arm interpolated `mut.evaluated` under a "survived" label, so a task whose test caught 1 of 10 mutants was rejected with "mutation score 0.10 (10 mutants survived)" — a claim the score in the same sentence disproves. This note is the entire operator-facing justification for rejecting a task, and it is fed back into the next worker's prompt.

**The cause was the type, not the format string.** A pluggable `FARM_MUTATION_CMD` need only print a trailing JSON line with a numeric `score` (`includes/farm.md:38`); `total` and `survived` are optional. But `MutationResult` declared all three as always present, so `parseMutationHookOutput` had to *invent* the missing ones (`?? 99`, `?? []`) purely to satisfy the type — and once invented, nothing downstream could tell a fabrication from a measurement. Every render had to guess, which is why four successive fixes each relocated the guess.

### The change

`evaluated` and `survivors` are **optional**, absent unless the producer reported them. The note states a survivor count only when there is one to state:

| hook JSON | before | after |
|---|---|---|
| `{"score":0.05,"total":10,"survived":[…9]}` | `10 mutants survived` | `(9/10 survived)` |
| `{"score":0.05}` | `(99 mutants survived)` | `score 0.05` — no count claimed |
| `{"score":0.1,"survived":[…9]}` | `(9/99 survived)` | `(9 survived)` — no invented denominator |
| `{"score":0.05,"total":10,"survived":[1,2,3]}` | `(3/10 survived)` | `(3/10 survived)` |

A count is read as reported when it is a number **or a numeric string**, so `"total":"10"` from a shell hook that quotes its numbers behaves exactly as before.

### Behaviour change to the escalation gate — measured, not asserted

The `evaluated >= 5` floor refuses to hard-reject on thin evidence. It previously substituted `99` for an unreported count, so a hook that never said how many mutants it ran could get a task rejected on no evidence at all. It now requires a reported count.

**27 hook shapes executed against both trees. Exactly four change**, all `escalate → warn`, all under one rule — *the producer stated no usable count*:

`{"score":0.05}` · `"total":null` · `"total":[10]` · `"total":1e400`

Nothing newly escalates. The built-in mutator is untouched.

### Review history — five rounds, each found real defects

| round | found |
|---|---|
| 1 | `survivors.length` empty on the hook path → `0/99`; a fabricated `DECISION-0029` citation in shipped source |
| 2 | deriving unconditionally discarded a hook's real survivor list (9 → 89); clamp ordering; `Math.round` load-bearing but untested |
| 3 | an invalid `total` cleared the escalation floor via the sentinel → `99/99 survived` on a run that evaluated zero |
| 4 | filtering non-string survivor ids silently shortened the count → `0/10 survived` while escalating *for* survivors |
| 5 | rejecting quoted numbers disabled the gate entirely for shell hooks; a mutant surviving behind a `?.` assertion that could not fail |

Three of those were defects in the *previous round's* fix, and three commit messages carried false claims. Both are recorded here rather than tidied away — the pattern is the useful part.

### Verification

- 9 files / **517 tests**; `tsc --noEmit` clean; `farm.js` rebuilt and byte-identical to a fresh build.
- **5 mutants this round, 5/5 killed**, including the round-5 survivor.
- Regression tests drive the repro inputs through the **real** `parseMutationHookOutput`. The `mutResult` helper builds `survivors.length === evaluated − killed` *algebraically* (verified 6/6 call sites), so it was structurally incapable of catching any of this.

Closes #525.


SUaDtL:
## Adversarial review found two BLOCKs. Both were real, both are fixed in `7bc1429`.

### 1. The fix regressed the hook path — worse than the bug it fixed

`mutationSurvivalNote` read `survivors.length`. But a pluggable `FARM_MUTATION_CMD` is only required to print a trailing JSON line with a numeric `score` (`includes/farm.md`) — `survived` and `total` are optional. On that path `parseMutationHookOutput` yields `survivors: []` with `evaluated` at its `99` fallback, which clears the `evaluated >= 5` floor, so the escalate arm *always* fires:

| hook output | old note | this PR, before the fix |
|---|---|---|
| `{"score":0.05}` (doc-minimal) | `score 0.05 (99 mutants survived)` | **`score 0.05 (0/99 survived)`** |
| `{"score":0.05,"total":10,"survived":9}` | `score 0.05 (10 mutants survived)` | **`score 0.05 (undefined/10 survived)`** |

It told the operator that **zero** mutants survived while rejecting the task **for having too many survivors** — the same self-contradiction #525 was filed about, moved to another path and made strictly worse. The old note was off by 5%; this was off by 100% and self-refuting.

It also falsified two claims in my own commit message: "the count and the score now agree" and "removing the class rather than the instance". On the hook path, neither was true — the warn arm's pre-existing weakness was *propagated into* the escalate arm rather than removed.

**Fix:** the count is now derived. `score` is `killed / evaluated`, so `evaluated * (1 - score)` is exact arithmetic over the two fields the contract *does* guarantee, and it equals `survivors.length` on the built-in path by construction. Clamped, since a hook's score is arbitrary operator input. `parseMutationHookOutput` now accepts `survived` only when it is genuinely an array of strings.

### 2. The previous commit cited a decision record that does not exist

It referenced `DECISION-0029` in `farm.ts` — which bundles into `farm.js`, so the dangling pointer would have **shipped** — and rested a rationale on it. The log ends at `DECISION-0028`. H-11 had blocked the attempt to write the record, and the citation went out anyway.

This is the same failure mode this repo already had to correct once by hand (DECISION-0014 vs commit `3902096`), and I reproduced it. The citation is removed; the reasoning stands on its own without claiming a record behind it. Whether the decision is logged is the maintainer's call — it needs `/ca:adr` to arm H-11, which is a routing gap worth its own issue, since `decision-variance` Phase 4 is *required* to append there.

### Verification after the fix

- `tsc --noEmit` clean; 9 files / **494 tests** pass; `farm.js` rebuilt.
- **6 new mutants, 6/6 killed** — including reverting the count to `survivors.length` (the regression above), both clamps, killed-vs-survived inverted, and both halves of the `survived` validation.
- New tests drive the doc-minimal hook shape end-to-end through `runTask`'s escalate arm, plus five `parseMutationHookOutput` shape cases.
- The reviewer separately confirmed: `farm.js` is byte-identical to a fresh esbuild 0.28.1 build, the union narrowing is sound (no `MutationHookFailure` can reach the formatter), and nothing in any plugin, workflow, hook, or site file parses these note strings.

SUaDtL:
## Round 2 (against `7bc1429`, the state round 1's fixes produced) — seven findings, all real, all fixed in `658c3b2`

Round 1's fixes were verified correct on the built-in path — a brute force over ~4.5M `(evaluated, killed)` pairs found 0 mismatches between the derivation and `survivors.length` — but they were wrong on the hook path they were written for.

### The derivation was over-applied (HIGH)

Round 1 correctly said `survivors.length` is empty on the hook path. I responded by **never** using it — which replaces a hook's *true* survivor count with an inferred one whenever the hook does report it:

| hook output | before this PR | round 1's fix | now |
|---|---|---|---|
| `{"score":0.1,"survived":[9 tags]}` | `9/99` | **`89/99`** | `9/99` |
| `{"score":85.2,"total":100,"survived":[15]}` | `15/100` | **`0/100`** | `15/100` |

`total` is optional, so the first row is legal output. That is #525's own defect pointed the other way — **the second time in this issue I fixed one direction by breaking the other.** The count now prefers the list and derives only when there is none.

### The clamp did not clamp (HIGH)

`min(evaluated, max(0, x))` applies the zero-floor and then lets a negative denominator drag the result back under it: `{"total":-5}` rendered `-5/-5 survived`. `max` now runs last. The test named "clamps … rather than rendering a negative survivor count" only exercised score-driven negativity, which `max` already handled — it did not pin the property in its own name.

### `Math.round` was load-bearing and untested (HIGH)

`floor` and `trunc` both passed the **entire suite**. They are not equivalent: at 12 evaluated / 5 killed — an ordinary size under the default `FARM_MUTATION_SAMPLE=15` — `floor` gives 6 where the truth is 7. Every fixture I had written happened to produce an integral product.

The previous commit called the product "exact arithmetic". It is not — `12 * (1 - 5/12)` is `6.999999999999999` — and believing it exact is exactly why the rounding went untested.

### The wrong field was hardened (MEDIUM)

Round 1's `survived` validation guarded a field with **no remaining runtime consumer**, while `total`/`evaluated` — the denominator the arithmetic divides by — stayed unchecked. `{"total":"abc"}` rendered `NaN/abc`, a regression this PR introduced by feeding an unvalidated field into arithmetic. Only a positive integer is accepted now.

### CHANGELOG (MEDIUM)

It claimed "the warn arm's output is unchanged", which the derivation made false on the hook path, and the derivation commit shipped no changelog at all. Both corrected — operators using `FARM_MUTATION_CMD` are now told plainly that a survivor count may be *inferred from the score* when their hook does not report one.

### Verification

- 9 files / **502 tests** pass; `tsc --noEmit` clean; `farm.js` byte-identical to a fresh esbuild build.
- **11 mutants over both files, 11/11 killed** — both rounding alternatives, both halves of prefer-vs-derive, the clamp order, the upper clamp, all three parts of the denominator guard, and the `total`-vs-`evaluated` precedence, which nothing had pinned.
- One mutant (clamp order) survived until I added a direct test: with the denominator now validated, the order is unreachable from either producer — but `mutationSurvivalNote` is exported and its signature admits a negative, so the contract is pinned rather than assumed redundant.

Issue #530, opened 2026-07-27
ca payload gate keys on the git tag, but installs key on the manifest version — untagged versions ship silently


SUaDtL:
The `ca` payload-version gate keys on the **git tag**; `claude plugin update` keys on the **manifest version string**. When those diverge, the gate reports "allowed" for a payload change that can never reach an installed user — the exact failure the gate exists to prevent.

Found live on 2026-07-27: two merged PRs of `ca` payload changes were invisible to the maintainer's own install.

## What happened

`plugins/ca/.claude-plugin/plugin.json` has read `2.9.1` since #313 (2026-07-24). No `v2.9.0` or `v2.9.1` tag was ever pushed — the newest `ca` tag is `v2.8.13`.

The gate (`ci.yml`, `[GATE ] | [CA ] | Payload version`) runs:

```
if [ "$(python3 .github/scripts/payload_scope.py --plugin plugins/ca --base "origin/$BASE")" = unchanged ]; then …
```

and then fails **only when `refs/tags/v<version>` already exists**. Since `v2.9.1` does not, every payload change riding `2.9.1` passes with *"payload changed on unpublished version — allowed"*.

Meanwhile the local install has:

```
~/.claude/plugins/cache/codearbiter/ca/2.8.11/
~/.claude/plugins/cache/codearbiter/ca/2.9.1/     <- populated, and stale
```

`claude plugin update` is a no-op while the manifest string is unchanged, so `2.9.1` in the cache stayed frozen at whatever the payload looked like when it was first installed. Verified: `grep -c DECISION_LOG_BASENAME` returns 3 in `plugins/ca/hooks/_hooklib.py` and 0 in the cached `2.9.1` copy.

## Why the gate's assumption fails

The guard's own header states the model correctly:

> `claude plugin update` is a NO-OP when the manifest version string is unchanged, so any change to the shipped payload that rides an already-published version silently never reaches installed users.

The bug is the definition of **"already-published"**. The gate equates it with *"a git tag exists"*. But a marketplace install populates its cache from the **manifest version**, with no tag involved. So a version can be *installed somewhere* while being *untagged*, and in that window the gate is inert against precisely the changes it is meant to catch.

Concretely, three `ca` payload changes shipped on `2.9.1` and reached no install:

- #524 — test-only, correctly out of payload scope, no impact.
- #527 — `skills/`, `includes/`, `agents/`: the coverage no-tooling exemption now requires a citation. Not in the cache.
[...]

Issue #528, opened 2026-07-27
decision-variance cannot record a decision: H-11 governs decision-log.md as if it were an ADR


SUaDtL:
The `decision-variance` skill is **required** to append every arbitration to `.codearbiter/decisions/decision-log.md`, and **cannot**, because H-11 blocks it. Any SMARTS arbitration outside an `/ca:adr` session therefore produces a decision that is made but never recorded.

Hit live on 2026-07-27 while running SMARTS on #525.

## The conflict

`decision-variance` Phase 4 (`core/surface/skills/decision-variance/SKILL.md`) requires: *"…then append the decision to the log per `includes/smarts/decision-log-format.md` — **immediately, never batched in memory**."*

H-11 (`plugins/ca/hooks/_hooklib.py:280`):

```python
DECISIONS_PATH_RE = re.compile(DECISIONS_DIR_RE + r"[\\/]+.+\.md$")
```

That matches **any `.md` anywhere under `.codearbiter/decisions/`** — including `decision-log.md`, which is not an ADR. `pre-write.py` / `pre-edit.py` then refuse the write unless a fresh `adr-authoring-active` marker is present, and `_bashguardlib.py` closes the shell flank (append redirects, `tee`, `sed -i`, …).

Only `decision-lifecycle` — the `/ca:adr` skill — arms that marker. `decision-variance` has no marker step and cannot legitimately add one: the marker's stated purpose is *"the mechanism enforcing ADRs only via `/adr`"*, and a second skill arming it would weaken the ADR guard in order to unblock an unrelated file.

## Why this is a classification bug, not a missing permission

`decision-log.md` is not an immutable ADR. It is an **append-only arbitration log** — the same shape as `overrides.log`, `triage.log`, `gate-events.log` and `sprint-log.md`, which are governed by **H-05**, not H-11. `decision-log-format.md` says so itself: *"Strictly append-only — no edits to a prior entry, ever… To supersede, append a new entry."* That is H-05's rule verbatim.

H-05 already has exactly the right machinery:

- `is_audit_log(rel)` — the append-only guard set (`_hooklib.py:298`)
- `is_tail_append(current, old, new)` — permits a verifiable tail-anchored pure append via Edit, refuses anything else (`_hooklib.py:305`)
- `pre-bash.py` permits an append redirect while blocking truncation, `tee`, `sed -i`, `rm`, `mv`

[...]

PR #531, opened 2026-07-27
chore(ca): release 2.10.0 so today's merged fixes actually ship

SUaDtL:
Remediation for the instance found while closing #528; the gate gap itself is **#530**.

### Why

#527 (skills/includes — the coverage exemption citation) and #529 (hooks — the H-11/H-05 arbitration-log reclassification) both merged onto an untagged `2.9.1`. `claude plugin update` is a no-op while the manifest version string is unchanged, so the populated `~/.claude/plugins/cache/codearbiter/ca/2.9.1/` stayed frozen at the pre-fix payload.

Verified: `grep -c DECISION_LOG_BASENAME` returns **3** in `plugins/ca/hooks/_hooklib.py` and **0** in the cached copy. The session that wrote the hook-guard fix was still being blocked by the pre-fix guard *after merging it*.

The version gate reported *"payload changed on unpublished version — allowed"*, because it keys on the git tag and `v2.9.1` was never pushed.

### Version derivation

The release window from the last **tag** (`v2.8.13`) classifies to a **minor** bump — 10 `feat`, 20 `fix`, 0 breaking — i.e. 2.9.0. But the manifest is already at 2.9.1, advanced past that by per-PR bump gates that never produced tags. Per the established rule that **the tag follows the manifest**, and since today's `ca`-scoped window is fix-only, `2.9.1 → 2.9.2` is the correct patch advance.

### Scope

- `plugins/ca/.claude-plugin/plugin.json` → 2.9.2
- `CHANGELOG.md` — `[Unreleased]` rolled into `[2.9.2]`, with the reason for the cut recorded
- `README.md` version badge

**No tag is pushed.** Publication stays a separate, explicitly authorized step.

Contracts verified: `build-surface --check` in sync, `sync-core --check` 48 × 3 byte-identical, docs contract clean.


SUaDtL:
### Correction: this is **2.10.0**, not 2.9.2

Caught while verifying the changelog roll, before merge.

I derived "patch" from **today's** `ca`-scoped commits. That was the wrong window — a release covers everything since the last *released* version, and `[Unreleased]` had accumulated a week of work that was never cut.

The real window carries **nine `feat` commits** since the 2.9.1 section, no breaking changes:

| | |
|---|---|
| #313 | governed Pi parity — `ca-pi` as a fourth sibling plugin |
| #471 | §6 intent routing + post-merge cleanup owner |
| #467 | doctor surfaces the MCP write-gap |
| #468 | published-tag drift detection |
| #486 | VOICE.md punctuation enforcement |
| #491 | ephemeral tool run is not a dependency |
| #493 | release lanes for `ca-sandbox` and `ca-pi` |
| #497 | `/ca:release` targets any of the four plugins |
| #503 | `/ca:review` reviews an inbound PR |

Highest precedence is **minor** → 2.9.1 → **2.10.0**.

Labelling a release that adds an entire plugin as a patch is a SemVer violation consumers act on. `/ca:release` Phase 2 exists to catch precisely this — it classifies the window and BLOCKS an explicit version that disagrees, rather than accepting the number a human proposed. I proposed 2.9.2 and did not run that classification against the real window until after this PR was open.

`plugin.json`, the CHANGELOG heading, and the README badge are all updated; no stale `2.9.2` remains anywhere.

Issue #575, opened 2026-08-01
H-22 lexical residuals: git restores a directory, package-manager 'install' false-blocks


SUaDtL:
Remaining declared residuals from the workstream-B adversary pass (MEDIUM-5 and the tail of MEDIUM-3). Both are lexical-reach limits of the H-22 shell flank, both are currently pinned as-is by tests, and neither is a regression.

## 1. `git` restores the whole directory without naming a basename

```
ALLOW  git checkout HEAD -- .codearbiter/
```

`git_restore_re` (the F5 leg) anchors on the bare basename, so restoring the containing DIRECTORY rewrites every enrolled file through git while naming none of them. This is the F5 leg's stated purpose ("a tracked worktree file can be rewritten through git itself") but not its coverage.

Same class, and honestly outside a lexical guard's reach — listed as scope, not as separate findings: `git stash pop`, `git reset --hard`, `git apply <patch>`.

## 2. A package manager's `install` subcommand is not coreutils `install`

```
BLOCK  pip install -r requirements.txt  # then read open-tasks.md
```

`install` is in `_STATE_WRITE_VERBS` for coreutils `install`, and the same-line window then reaches a basename mentioned in a trailing comment. This is the declared T-08b same-line residual and is currently **pinned as still-blocking** in `test_write_verb_window_does_not_cross_a_newline`, so the boundary is visible rather than silent.

Narrowing it means distinguishing `pip|npm|cargo|apt|brew install` from `install -m 644 src dst` — a preceding-token check, not a smarter parser.

## Not in scope

The multi-line reach of this window WAS fixed (#564 remediation): the window is now `[^|;&\n]*`, so a write verb on one line of a commit body and a basename several lines later no longer form a match. Before that fix, 1 of this branch's 49 commit messages would have been refused by the guard it installs; after, 0.

Refs #564.

Issue #574, opened 2026-08-01
H-05/H-11/H-18 have no interpreter leg at all — an inline-code one-liner walks past three guards


SUaDtL:
Found by the workstream-B adversary pass's cross-guard parity probe (HIGH-5, the non-H-22 half) and independently confirmed by reading the source.

## The gap

H-22's shell flank and H-19's gate-marker check each carry an interpreter leg. The three older protected-file guards carry **none**:

```
$ grep -c "python3?|node|perl|ruby|sh" core/pysrc/_hooklib.py
0
```

`LOG_TRUNC_RE`/`LOG_DESTROY_RE` (H-05, `overrides.log`), `DECISIONS_*` (H-11, decisions/ADRs) and `CONTEXT_*` (H-18, `CONTEXT.md`) match redirects and write verbs only. So an interpreter one-liner reaches all three:

| shape | H-22 open-tasks.md | H-05 overrides.log | H-11 decisions/ADR | H-18 CONTEXT.md |
|---|---|---|---|---|
| `python3 -c` | BLOCK | ALLOW | ALLOW | ALLOW |
| `py -c` | BLOCK (as of #564 remediation) | ALLOW | ALLOW | ALLOW |
| `pwsh -Command` | BLOCK (as of #564 remediation) | ALLOW | ALLOW | ALLOW |

The H-22 column is current after the remediation commit on `sprint/portable-release-and-protected-state`; the other three are unchanged.

## Scope note

This is **pre-existing** and structural — not a regression from #564. It was left out of that sprint's remediation deliberately, to avoid widening scope into three guards the sprint did not touch.

## The fix is cheap

`_bashguardlib._INTERP_TOKENS` and `_INTERP_INLINE_CODE` already exist as shared constants (added in the #564 remediation). Adding one `interp_re` leg per guard, built from those, is the same edit three times.

Decide per guard whether the inline-code requirement applies: for H-22 it is load-bearing (a sanctioned helper is invoked as `python3 helper.py <args>` and matching the bare token blocked it). H-05/H-11/H-18 have no equivalent sanctioned interpreter caller, so they may want the broader token-only match H-19 keeps.

Refs #564, #237, #335.

Issue #573, opened 2026-08-01
Archival sweep produces a diff commit-gate unstages as scope creep (ADR-0008 conflict)


SUaDtL:
Found by the workstream-B adversary pass (MEDIUM-2) on `sprint/portable-release-and-protected-state`, independently reproduced.

## The conflict

`/ca:standup` step 6 (the per-item archival sweep, T-61..T-67) removes a done entry from `open-tasks.md`. `classify_board_diff` has no archive class and returns `False` on any line removal:

```
archive (removal)  -> False
done-flip control  -> True
```

`commit-gate/SKILL.md:97` (ADR-0008) then applies:

> Any other `open-tasks.md` change — a reworded or **deleted** entry, or an arbitrary content edit — does not classify as a transition and still flags as scope creep.

So every archive the sanctioned sweep performs produces a diff the sanctioned commit path unstages. The feature is reachable but not committable without `/ca:override`.

## Why this is filed rather than fixed

Making the sweep work means adding a fourth clean-transition class (archive-removal) to `classify_board_diff`. ADR-0008 **explicitly** rules that a deleted entry is scope creep, so that is an amendment to an accepted decision, not a bug fix — and ADRs are user-attributed, authored only via `/ca:adr`. It needs a ruling before code.

## The shape a fix would take

An archive-removal class would need to be narrow enough not to re-admit the deletion ADR-0008 is protecting against: the removed block must be `[x]`, must carry a `(done YYYY-MM-DD)` stamp, must appear verbatim in `done-tasks.md` in the same commit, and nothing else may change. That last clause is what distinguishes it from an agent quietly dropping an inconvenient task.

Note `test_done_flip_retained` exists to prove the done-flip stays RETAINED; nothing equivalent exists for archive.

Refs #564, ADR-0008.

SUaDtL:
## Corrected framing

[...]

PR #576, opened 2026-08-01
feat(release): a portable release fixture and the protected-state machinery it needs (#563, #564)

### LINKED ISSUE #563: release skill ships codeArbiter's own release process to every consumer repo
`/ca:release` is the one shipped skill that is not portable. It encodes this repository's release mechanics as skill logic rather than reading them from project state, so a repo that installs codeArbiter gets a lane that cannot execute.

> **Scope note.** This body was rewritten 2026-07-30 after design work materially widened the change. The original body described only the release-skill fix. See [the earlier correction comment](#issuecomment-5138965100) for the measurement errors it contained, which are fixed below.

**Spec:** `.codearbiter/specs/release-portable-fixture.md` (rev 3, after two adversarial review passes)
**Decisions:** DECISION-0034 in `.codearbiter/decisions/decision-log.md` (supersedes DECISION-0033)

## The hard failure

Pre-flight's first resolution step is:

```
TAG_PREFIX=$(python3 .github/scripts/_releaselib.py tag-prefix $TARGET)
```

That helper is **not in the plugin payload**. Neither are the other two files the skill depends on:

| referenced by the skill | actual location | under `plugins/ca/`? |
| --- | --- | --- |
| `_releaselib.py` | `.github/scripts/_releaselib.py` | no |
| `build-host-packages.py` | `tools/build-host-packages.py` | no |
| `published-tags.json` | `.github/published-tags.json` | no |

In a consumer repo those paths do not exist. The likely agent response is not a clean error but an improvised substitute, which in a gate skill is the worse outcome: the whole point of `release` being the single permitted path to a version tag is that it does not get improvised.

## Measurement rule

Contamination counts are pattern-dependent, and the original body quoted a number without stating its tokenizer. The rule, not the count, is what matters: **a reference is contaminating when it names a path belonging to this repository that the skill executes or reads.**

Bare substring matching is explicitly rejected. `plugins/ca/skills/context-creation/SKILL.md:49` has a scout read `.github/workflows/`, `.gitlab-ci.yml`, and `Jenkinsfile` — these describe the *consumer's* repository and are correct as written. A naive guard would flag them forever.

Under the reference-form rule, **three** shipped skills are contaminated:

- `release` — heavily, and the subject of this issue
- `subagent-driven-development:45` — dispatches `tools/farm.js`, which ships at `plugins/ca/tools/farm.js`, so the bare path resolves in neither form
- `decision-lifecycle:70` — names `.github/scripts/check_adr_identity.py`, a CI-only script that is not shipped

## When it happened

Not at the four-plugin change, which is the intuitive suspect:

```
6a45173   Phase 7a: relocate plugin into plugins/ca/            (portable, 51 lines)
261374b   feat(release): publish the GitHub Release in Phase 3
c12b1a3   Release-skill hardening + test-debt paydown (#125)    <-- inflection
29ab107   fix(release): harden /ca:release with tested guards
3d5f6c9   feat(ci): detect published-tag drift (#468)
c20c2d0   feat(release): targets any of the four plugins (#497)
```

Independently verified: `6a45173` shipped a 51-line skill with zero repo-path references, and `c12b1a3` is an ancestor of `c20c2d0`. Portability was lost five commits before any multi-plugin work; #497 inherited an already-repo-specific skill.

None of those commits were careless. The original portable version resolved the last tag with bare `git describe --tags --abbrev=0`, which the current skill correctly condemns as the thing that silently bases a release on a sibling plugin's tag. Every hardening commit fixed a real failure. The defect is that each fix was written as a hardcoded fact about this repository instead of a parameter read from project state.

## The maintainer path is healthy

This is a portability defect, not an active breakage. All four subcommands the skill invokes are implemented and run (`tag-prefix ca` returns `v`, `last-tag v` returns `v2.8.13`), CI is green on main, and `sync-core --check` passes across 53 files x 3 plugins.

## Design

Full detail in the spec. In summary:

1. **Split `_releaselib`** into portable *mechanism* — which moves to `core/pysrc/` and generates byte-identically into all three governance plugins via the existing ADR-0011 mechanism — and repo-specific *data*, which moves to project state. Repo-specific defaults inside the mechanism (`MERGE_READINESS_CHECK`, the `last_tag_select` prefix, `select_release_target`'s target list) become required parameters.
2. **`.codearbiter/release-targets.md`** carries one sub-block per target with a defined grammar and an explicit parser contract, parsed stdlib-only per ADR-0004 using the HTML-comment delimiter convention `_scopelib` already recognizes. It lives outside `CONTEXT.md` on **context economy**: `CONTEXT.md` is read every session, release configuration only when tagging.
3. **Pre-tag commands are declared per row and check-only** (DECISION-0034). They assert and exit non-zero on drift; they may never mutate the tree, and the clean-tree assertion is unconditional. Reconciliation stays a separate operator action routed through `commit-gate`.
4. **A new protected-write class** guards the declaration file, since it is the only *executable* file in `.codearbiter/` and no existing class fits — `context` admits any write preserving `arbiter: enabled` frontmatter, `marker` blocks outright, `audit` is append-only, and `decisions` requires a marker only `/adr` mints.
5. **Onboarding owns elicitation.** `context-creation` scouts and writes full rows; `decompose` elicits intent only, since it runs before any manifest or tag exists. Release-time detection is a back-fill path requiring explicit confirmation.
6. **CI reads the same declared source**, and target selection becomes **name-keyed rather than positional** — today `select_release_target` zips confirmations positionally against `RELEASE_TARGETS` while `release.yml` passes them in hardcoded order, so moving row order into an editable file would let a reorder remap which input drives which `contents: write` publisher.
7. **`.github/scripts/_releaselib.py` becomes a permanent, data-free shim** rather than being deleted, because six CI sites shell out to that path (`release.yml:135,171` and `publish-release/action.yml:125,164,180,228`).

## Acceptance

42 criteria across six slices in the spec. The load-bearing ones:

- [ ] A guard scans `core/surface/skills/**` and fails on any reference naming a this-repo path the skill executes or reads, by the stated rule, permitting scout scan-target patterns
- [ ] All three contaminated skills fixed, not only `release`
- [ ] A consumer repo with one artifact runs `/ca:release` end to end with no file outside the plugin payload
- [ ] This repo still releases all four plugins with per-series isolation and payload scoping intact, proven by a resolution trace against a recorded pre-change run covering `ca` and `ca-pi`
- [ ] No repo-local variant of the skill exists
- [ ] No commit leaves a CI consumer broken; the migration ordering is stated and pinned
- [ ] `commands/release.md` matches the skill it routes to
- [ ] The docs-site guide distinguishes the general lane from this repo's configuration

## Review provenance

Two adversarial passes against the spec. Pass 1 found the criteria set literally unsatisfiable (the portability guard forbade the `.github/` path the tag-provenance hard rule requires), the slice ordering CI-breaking, the provenance drift trigger decorative, the golden test's oracle circular, and the positional target selection a reorder-to-mispublish hazard. Pass 2 verified 11 of 15 repairs sound and found four defects, two introduced by the repairs themselves. A third narrow pass on the new protected-write class is in progress.

### LINKED ISSUE #564: protected-state machinery: a registry-backed write guard so helpers, not inference, mutate .codearbiter/ state
Project-state files under `.codearbiter/` are freely agent-writable. Some already have a sanctioned helper — `taskwrite.py` says of itself, "the ONLY blessed way to write `.codearbiter/open-tasks.md`" — but nothing *enforces* that path, so an agent composing board markdown directly is indistinguishable from the helper's output until something downstream disagrees.

This issue builds the guard that makes the helper the only route, as generic registry-based machinery rather than a per-file special case.

**Spec:** `.codearbiter/specs/portable-release-and-protected-state.md` (workstream B)
**Plan:** `.codearbiter/plans/portable-release-and-protected-state.md`
**Related:** #563 (release portability — `release-targets.md` is the first consumer)
**Decisions:** DECISION-0034; sprint-log D-1/D-2 closure and B2 policy entries, 2026-07-31

## Why a registry with per-entry policy, not a uniform guard

An adversarial pass killed the first design, which marker-gated all consumers uniformly. The three files need materially different write semantics:

| file | policy | rationale |
| --- | --- | --- |
| `release-targets.md` | `marker-gated` | No helper exists; `context-creation` composes it with the Write tool. The H-11 marker pattern fits. |
| `open-tasks.md` | `helper-only` | Hard block on Write/Edit/shell naming the file, with **no marker path**. |
| `done-tasks.md` | `append-only` | Mutation only through the archive verb. |

`open-tasks.md` must *not* be marker-gated. `taskwrite.py` is Python file I/O invoked via Bash whose argv never lexically names the file, so it is **invisible to all three hook flanks by construction** — a marker would add nothing for the helper while *admitting* an agent that composes board markdown under that marker. That is the exact inversion of the goal.

The policy enum ships in the first slice even though only `marker-gated` is exercised there. A marker-only schema would need a breaking rebuild when the board files enroll.

## All three flanks, or it is not a guard

`pre-write.py` is one door. The class must also register on `pre-edit.py`'s `classify_protected` dispatch and on a `_bashguardlib` redirect + write-verb regex pair mirroring `CONTEXT_REDIRECT_RE` / `CONTEXT_WRITE_RE` (lines 355-356, checked at 1011). A one-flank implementation passes a Write-door test while `echo '...' >> <file>` still lands.

Two regression tests are obligations, not enrolment-time discoveries:

- `git add open-tasks.md` must **pass** — `commit-gate` Phase 7 runs exactly that (`commit-gate/SKILL.md:113`). If the mirrored verb set drifts to include git verbs, commit-gate blocks itself on every retained board flip.
- `taskwrite add -- "fix open-tasks.md schema"` must **pass** — the filename appears in argv as data.

The shell-indirection residual (`f=...; sed -i "$f"`, novel `python -c` spellings) is the same accepted ADR-0010 cooperative-attestation residual `CONTEXT.md` already carries, and must be named in the accompanying ADR rather than left implied. The marker's value is audit friction, not authorization.

## Closes D-1 and D-2

**D-1** closes as fact rather than decision. `taskwrite.py:2` literally says "resolves D-1"; the `open-questions.md` text predates the writer and is stale. What D-1 actually left open is that the sanctioned path is not enforced — which is this issue.

**D-2** needs a new helper verb before it can be closed at all. `taskwrite.py`'s verbs are exactly `add`/`start`/`done` with no removal, and the routing table says "never delete to complete." With both board files enrolled, a model Edit performing the archival move is blocked by this issue's own guard.

- **`taskwrite archive <ID> [--date]`** — **per item**, append to `done-tasks.md` first, then remove from `open-tasks.md`. Batch ordering is unsafe: appending all N then removing all N duplicates every item if interrupted between phases, and the reverse order loses records.
- **Owner: `/ca:standup`**, whose per-action confirmation and never-destructive-without-a-yes contract is verbatim what D-2 asked for. Per-item ordering maps its confirmation 1:1 onto helper calls.
- **Cutoff: done > 14 days**, a named constant tested against an injected date. Logged `confidence: low` — there is no evidence behind 14 over 7 or 30.
- **Undated `[x]` items** archive only under explicit per-item confirmation; both `taskwrite done` and the ADR-0008 classifier enforce `(done ...)` stamps, so an undated entry is legacy or override-era.

## The conversion set is two surfaces, not the whole board

A verified inventory replaced an earlier keyword estimate of seven or eight writers:

- **Already helper-routed, no change needed** — `harvest.md:51-52` promotes work via `{{CMD:task}} add`, so `commit-gate`, `tdd`, `brainstorming`, and `writing-plans` all already write through the helper.
- **Not writers at all** — `using-git-worktrees:34` explicitly says *not* the backlog; `standup.md:44` says "the board is never mutated here"; `reference-map.md` is routing.
- **Hook-layer, helper-path, unaffected** — `init-codearbiter.py:96` creates the file; `taskwrite.py` itself. Verified read-only: `boardsync.py`, `session-start.py`, `statusline.py`, `_arbiterstatelib.py`.
- **Needing real work** — `debug/SKILL.md:80` does a direct append carrying a rationale sub-bullet that `taskwrite add` cannot yet express (conversion **plus** a helper extension), and `context-creation/SKILL.md:103` populates the board with the Write tool during scaffolding (needs a helper route or a declared scaffold-time exemption).

## ADR-0008 composes cleanly

Verified by trace: `/ca:task done X` → `taskwrite.py` stamps `(done YYYY-MM-DD)` → the Bash flank sees no lexical filename and passes → `commit-gate` Phase 6 `classify_board_diff` sees a clean done-flip (the stamp is *required* by the classifier) → RETAINED → Phase 7 stages by explicit path. The write-time guard and the commit-time classifier act on different objects — a tool call versus a staged diff — so they are complementary, not double-gating.

## Circularity

The board is how this work is tracked, so a wrong guard could block recording the fix. Under `helper-only` the guard **cannot** block `/ca:task`, because the lexical flanks never see the helper's write. That structural fact, not enrolment sequencing, is the load-bearing mitigation, and it gets a pinned test. The residual is an overbroad flank regex blocking unrelated Bash, where `/ca:override` remains available and logged.

## Acceptance

27 criteria (B-01 … B-27) in the spec. Headline set:

- [ ] Registry carries the policy enum from the first slice
- [ ] All three flanks registered; a one-flank implementation fails the test set
- [ ] `git add open-tasks.md` and filename-as-argv-data both pass; `tee` and `>>` block
- [ ] `taskwrite.py` still succeeds with enrolment live
- [ ] `taskwrite archive` is per-item and rerun-safe; interruption causes neither duplication nor loss
- [ ] `debug` and `context-creation` converted; `taskwrite add` extended for the rationale sub-bullet
- [ ] A done-flip still classifies RETAINED after enrolment
- [ ] An ADR records the executable-input boundary and names the ADR-0010 residual

SUaDtL:
Closes #563. Closes #564.

Two workstreams, 95 planned tasks, all ACCEPTED. 50 commits.

## A — `/ca:release` becomes a portable fixture

The release lane held this repository's four plugins as skill logic. It now holds none: every fact it needs — tag prefix, manifests, changelog, payload scope, pre-tag checks — is read from a project's declared `.codearbiter/release-targets.md`. One skill serves both a four-plugin monorepo and a single-artifact consumer repo, with no repo-local variant.

- `_releaselib.py` carries no fact about this repository, enforced by a denylist test.
- `run-pre-tag` runs the declared checks in order, stops at the first non-zero, and asserts nothing changed — digesting the **content** of every changed path, not porcelain lines. Drift and mutation get distinct exit codes.
- `row_assertions()` is one derivation behind A-3.1..3.5, so the five assertions cannot drift apart.
- The release workflow and the version gate read the declared file; no tag-prefix literals remain in CI.
- `check_skill_portability.py` (new) fails a shipped skill that names a this-repo path it executes or reads — a consumer does not have `.github/scripts/`.

A house rule runs through it: **never fold "I could not compare" into "the answer is no."** `semver-greater` and `check-manifests` return distinct codes for the two.

## B — a generic protected-state registry

`_protectedstatelib.py` is a path→policy registry (`marker-gated` / `helper-only` / `append-only`) enforced across all three hook flanks — `pre-write.py`, `pre-edit.py`, and `_bashguardlib.py`'s shell check. All three planned consumers are enrolled, one per policy: `release-targets.md`, `open-tasks.md`, `done-tasks.md`.

`taskwrite.py` gains an `archive` verb and `/ca:standup` a per-item archival sweep — one confirmation per item, one helper call per item, `done-tasks.md` written before `open-tasks.md` so an interrupted sweep is recoverable.

Governed by ADR-0024 (protected state is a declared executable-input boundary with cooperative, friction-grade markers).

## Adversary passes

One at Opus/medium per workstream. Every HIGH remediated in `1864c33`; the interesting part is that four of the five were introduced by this sprint's own preceding slices, and every one sat in a **seam** — how a caller wires mechanisms together — that the green mechanical suites structurally could not see.

| # | Defect | Fix |
|---|---|---|
| HIGH-1 | The interpreter leg matched the interpreter token alone, which is the shape of the **sanctioned** call. Every `/ca:task` invocation naming an enrolled file blocked — and for `done`/`archive` on an ID-less task the title *is* the target, so no reword existed and no sanctioned route remained. | Require an inline-code switch: running a file and passing the name as argv is data; executing a string that names it is a write. |
| HIGH-2 | `archive_transform` moved the top line only, orphaning `Desc`/`Boundaries` onto the *next* task — silently rewriting a security scope. | Move the whole block, boundary rule taken from `parse_board` itself. |
| HIGH-3 | Removal was text-equality over every line, so two identical done entries collapsed to one. A record destroyed at rc=0. | Index-based removal. |
| HIGH-4 | A failed read of the append-only archive became `""`, and the write replaces the file wholesale — any transient `OSError` wiped every historical record. | Split ABSENT from UNREADABLE; refuse at a distinct exit code before either write. |
| HIGH-5 | The interpreter list missed `py` — *the* Windows Python launcher — and PowerShell entirely, so the leg was bypassable on this repo's primary dev host. | One shared alternation, widened. |

MEDIUM-1/3/4/5 also fixed. MEDIUM-1 deserves a note: the test that *claimed* to pin archive write-ordering compared a string against itself, and a reversed-order mutant survived the entire suite. It now fails the second write and checks what survived on disk.

Deferred with reasons, not dropped: #573 (archive vs ADR-0008 — needs a ruling, not a fix), #574 (H-05/H-11/H-18 have no interpreter leg; pre-existing, three untouched guards), #575 (lexical residuals).

## Verification

All suites green on the final commit: every `.github/scripts/test_*.py` module, the site suite (47 files / 496 tests), and `build-surface --check` (claude, codex, pi in sync). The one local failure is a pre-existing `ca-pi` flake in a package this branch does not touch, filed as #580. **The proof gate is red, deliberately — see "Known state on merge" below.** `check-plugin-refs`, `check_adr_identity`, `check_skill_portability`, `check_command_catalog`, `check_docs_contract` all clean.

Every new test was confirmed to die to a mutant, **with the mutation asserted to have applied** — a rule this sprint earned when a heredoc escaping bug made a "surviving" mutant a no-op.

https://claude.ai/code/session_01C8oamhuiaBjZnauWRGjb86

---

## Review remediation (added after the original description)

**CodeRabbit** (ASSERTIVE profile) returned 69 findings. All **24 Major** are fixed — ~12 distinct, since `core/pysrc/X` and its two vendored copies each flag separately. Highlights:

- `task_block` diverged from `parse_board`'s closing rule, **re-opening HIGH-2** — indented continuation prose orphaned sub-bullets onto the following task.
- `_archive` parsed the board one line at a time, so every `Task` carried `lineno == 1` and HIGH-3's index-based location was **dead code** running on its fallback.
- `git_restore_re` still crossed newlines — the same defect fixed in `write_re` two lines above it.
- A `delete` op reached the marker-gated arm, making a self-mintable friction-grade marker a **delete capability** over protected state.
- `run-pre-tag`'s mutation check was blind to a command that REVERTED an edit.
- `confirmation_path` collided `a.b` with `a_b`, so confirming one confirmed the other.
- The `python3 X || python X` convention **discarded helper exit codes**: `||` re-runs on any non-zero, so `run-pre-tag`'s 5/6 verdicts triggered a second full execution of the project's declared pre-tag commands and reported the second run's code.

**Two blind agent-judgment exercises** (runs 14 and 15) drove the shipped skill as an operator's agent would. Nine HIGHs fixed:

- The lane forbade reading the declared file by eye but shipped readers for only 2 of 13 fields, so following it *required* the forbidden path. Added `show-row` and `show-row --field`.
- `$PAYLOAD` is "payload minus payload-exclude", which `git log --` cannot express; added `payload-pathspec`.
- The changelog roll had no floor when a manifest runs ahead of the last tag, so a release would re-publish already-shipped entries. Now STOPs.
- `show-row`'s first cut emitted unquoted values and the skill documented `eval`-ing them — which **executed declared shell** (`rebuild: cd x && npm run build` → `REBUILD=cd` plus the command `x`) while still exiting 0. Values are `shlex.quote`d; `--field` needs no eval.
- `MAJOR.MINOR.PATCH` was a literal in 7 tag/push/release commands and never assigned; `$VERSION` now is.
- The clean-tree gate could never pass, because this layer's own hooks append to `gate-events.log` on the very commands the lane runs.

**Deferred, filed not dropped:** #573 (archive vs ADR-0008 — needs a ruling), #574 (H-05/H-11/H-18 have no interpreter leg), #575 (lexical residuals), #577 (repo-wide `||` convention), #578 (45 minor/trivial), #579 (remaining exercise findings).

**Three more blind exercises (runs 16, 17, 18)** drove the shipped skill after the ones above. Six more HIGHs, all fixed:

- Every helper invocation broke on a plugin root containing a **space** — `"$PY" {{PLUGIN_ROOT}}/hooks/<script>`, interpreter quoted, path beside it not. `C:\Users\First Last\.claude\plugins\` is an ordinary Windows install; on such a host every step fails at once.
- Phase 1 step 7 asserted clean-tree with the **bare** `git status --porcelain` while calling it "the same state Pre-flight required" — which Pre-flight's own HIGH-rated paragraph proves cannot pass there.
- Step 6c's own stated remedy (`releasehash.py record`) wrote a marker the step-7 gate then refused — a hard BLOCK at the last gate of Phase 1, masked here only by a `.gitignore` line and **not** masked for a Back-fill consumer, which is exactly who the lane serves.
- The clean-tree pathspec used a cwd-relative `.`, and a declared `rebuild` commonly **begins with `cd`**. From the subdirectory that leaves you in, three gates fail silently: a false "nothing to release" STOP on a 95-commit window, an artifact-freshness check that passes without resolving the artifact, and a dirty tree reading as clean.
- Step 3 told the operator that amending or rebasing a footer-less commit was "fine, since nothing in this window has been tagged or pushed". **38 of the 57** footer-less commits are already on `origin/main` — the sentence prescribed a force-push over published history that this same file forbids twice.

Run 18 then attacked all four of the earlier fixes directly and they held, including from a path containing a space.

**Known state on merge:** `check_skill_proof_fresh.py` is **red**, and that is a decision rather than an oversight. Run 18 exercised the skill at `33e82a6a` and found the force-push instruction above; fixing it moved the skill one commit past the hash the exercise covered. The honest options were to record a proof for bytes an agent actually read and leave the gate red, or to ship a destructive instruction to keep the gate green. `proof_current` is `false`, `post_exercise_delta` in the artifact names exactly what changed after the exercise and why, and the gate blocks the next `/ca:release` for `ca` until run 19 runs against the final bytes.

The gate was already red before this session — stale since `8f101403`, still recording run 13. It now records run 18 with a named one-commit delta, which is strictly better than red-and-five-commits-stale. It is a declared `pre-tag` row, not a CI job, so it does not gate this merge.

**Deferred, filed not dropped:** #573 (archive vs ADR-0008 — needs a ruling), #574 (H-05/H-11/H-18 have no interpreter leg), #575 (lexical residuals), #577 (repo-wide `||` convention), #578 (45 minor/trivial), #579 (earlier exercise findings), #580 (`ca-pi` bridge flake), #583 (run-16 non-HIGHs), #584 (run-17 non-HIGHs), #585 (run-18: Back-fill's first release cannot clear step 3, plus four MEDIUMs).


SUaDtL:
## Disposition of the CodeRabbit review threads

Resolving the 60 open threads so `required_conversation_resolution` clears. Recording the disposition here rather than resolving silently, because "resolved" and "addressed" are not the same word and the distinction should survive this PR.

**All 24 Major findings were fixed**, not dismissed — commits are in the PR description. Roughly 12 are distinct: `core/pysrc/X` and its two vendored copies each flag separately, so fixing the source and re-syncing clears three threads at once. The ones worth remembering:

- `task_block` diverged from `parse_board`'s closing rule and **re-opened HIGH-2** — a fix from earlier in this same sprint.
- `_archive` parsed the board one line at a time, so every `Task` carried `lineno == 1` and HIGH-3's index-based location was **dead code** running on its fallback.
- `git_restore_re` still crossed newlines — the identical defect fixed in `write_re` two lines above it.
- A `delete` op reached the marker-gated arm, making a self-mintable friction-grade marker a **delete capability** over protected state.
- `confirmation_path` collided `a.b` with `a_b`, so confirming one confirmed the other.
- The `python3 X || python X` convention **discarded helper exit codes**: `||` re-runs on any non-zero, so `run-pre-tag`'s 5/6 verdicts triggered a second full execution of the project's declared pre-tag commands and reported the second run's code.

**45 minor and trivial findings are filed as #578** — stale prose, table formatting, markdownlint items. Real, none load-bearing, and batching them keeps this PR reviewable rather than dropping them.

**28 of the 60 threads are already `outdated`** — the lines they point at have been rewritten since, in most cases by the fix for the finding itself.

Nothing here is being closed as wrong. If a thread below turns out to name something neither fixed nor in #578, it should be reopened — file it against #578 and it will be picked up.

Issue #578, opened 2026-08-01
CodeRabbit #576: the 45 minor/trivial findings not taken in that PR


SUaDtL:
The 24 Major findings from CodeRabbit's review of #576 were fixed in that PR (plus the two follow-ups #577 and this one). These are the deferred remainder: 35 minor + 10 trivial. All real, none load-bearing; batched so #576 stayed reviewable.

Full thread: https://github.com/arbiterForge/codeArbiter/pull/576

## Stale prose and governance records

- `_protectedstatelib.py`: the "registry ships EMPTY" claim is stale in **six** places now that three consumers are enrolled.
- `_bashguardlib.py`: the same stale empty-registry comment (both vendored copies).
- `specs/release-portable-fixture.md`: header still says `awaiting approval`; the Phase 1 gate cleared.
- `plans/portable-release-and-protected-state.md`: prose blocks sit between table rows without a blank line, so markdown parses them as one-cell rows; and the coverage proof disagrees with the ledger on A-6.7/T-77 and A-6.8/T-79.
- `reports/agent-lane-proof.json`: the stopping-rule narrative says eight exercises, the campaign ran 13; `verification.as_of` still names run 8 / commit 5192ef7.
- `security-controls.md:513`: `#175` at line start trips markdownlint MD018.
- Root `CHANGELOG.md`: one statement needs correcting; README version badge lags `package.json`.
- `debug/SKILL.md`: missing relative pronoun.

## Docstring and API surface

- `_releaselib.py`: Public API list omits four exported functions; `provenance_trigger_paths` docstring omits `generated_manifest`; `--targets-file` doc is wrong for two subcommands; a blank `rebuild:` is reported as a declared command; a pair with an empty name returns `""`, outside the documented label set.
- `_releaselib.py`: add `ValueTooLongError` to the declared-exceptions list; add a timeout to the git probe.
- `.github/scripts/_releaselib.py`: `select_release_target_by_name` is either public API or should not be re-exported.

## Robustness

- `check_skill_proof_fresh.py`: validate the document type before `.get()` — valid JSON of `[]` or `null` raises `AttributeError`.
- `payload_version_gate.py`: two payloads with the same basename overwrite each other silently; catch `ReleaseTargetsError` around `tag_prefixes()`.
- `_releaselib.py`: any `OSError` maps to `AbsentBlockError`, so a permissions failure can trigger the back-fill lane — same "unreadable is not absent" class as the `taskwrite` fix in #576.
[...]

Issue #577, opened 2026-08-01
The `python3 X || python X` interpreter convention masks helper exit codes repo-wide


SUaDtL:
Found by CodeRabbit on #576 and confirmed. Fixed in the `release` skill there; this issue covers the rest of the repo.

## The defect

The documented cross-host interpreter convention is:

```
python3 "<script>" <args> || python "<script>" <args>
```

`||` branches on the **exit code**, so it cannot distinguish *"there is no `python3`"* from *"the helper ran and told you something"*. Any non-zero exit re-runs the whole command under `python` and reports the **second** run's code.

That is fine for a helper whose only meaningful answer is 0/non-zero. It is actively wrong for one whose exit codes carry information — and this repo now has several:

| Helper | Codes |
|---|---|
| `_releaselib.py run-pre-tag` | 5 = a check reported drift, 6 = a check mutated the tree |
| `_releaselib.py semver-greater` | 1 = not greater, 2 = could not compare |
| `_releaselib.py check-manifests` | 1 = a path disagrees, 2 = a manifest could not be parsed |
| `releasehash.py check` | 1 = commands changed, 2 = never confirmed |
| `taskwrite.py archive` | 3 = the append-only archive could not be read |

For `run-pre-tag` the consequence is concrete: a drift result re-executes the project's declared pre-tag commands a second time and discards the first verdict.

## The fix

Resolve the interpreter once, by presence, then use it:

```sh
PY=python3; command -v python3 >/dev/null 2>&1 || PY=python
"$PY" "<script>" <args>
```

`command -v` tests whether the interpreter EXISTS, which is the actual condition the fallback is for.

## Scope

Already fixed in `core/surface/skills/release/SKILL.md` (#576). Remaining: every other skill and command surface using the `||` spelling. Worth a sweep with a guard afterwards, so the pattern cannot come back.

```
grep -rn '|| python' core/surface/
```

Note the same `||` shape in `hooks.json` is **fine** and should be left alone: those hooks are pass/fail, with no exit-code vocabulary to lose.

Refs #563, #576.

PR #587, opened 2026-08-01
chore(coderabbit): stop re-reviewing the vendored plugin trees

SUaDtL:
Adds a `.coderabbit.yaml` so CodeRabbit stops reporting the same defect once per vendored copy.

## The problem, measured

This repo vendors one source of truth into three host plugins. `tools/sync-core.py` copies
`core/pysrc/*.py` byte-identically into each plugin's `hooks/` directory; `tools/build-surface.py`
renders `core/surface/` into each plugin's commands, skills, routines, agents and includes per
`core/hosts.json`. CI fails the build if any copy drifts, so a defect in `core/` is mechanically
present in every copy.

Unconfigured, CodeRabbit reviews each copy independently. On #576:

| | |
|---|---|
| Inline comments | **102** |
| On `plugins/` (derived trees) | **49 (48%)** |
| `SKILL.md` flagged | **40 times** |
| `_releaselib.py` flagged | **15 times** |

Replaying #576's comment paths against this config: **48 of 102 filtered, 54 still reviewed.**
That volume is what forced the deferrals in #577 and #578.

## Design choices, and what they cost

**Negation patterns only.** It is undocumented whether a positive pattern flips `path_filters`
into allowlist mode and silences every path it does not name. Guessing wrong loses review coverage
entirely, so nothing here is a positive pattern. The cost: the three authored `_host.py` files lose
review, because re-including them would require a positive pattern. Three small files traded
against 55 test files.

**Nothing filtered that was not proven derived.** `plugins/ca-sandbox/**` is absent from
`core/hosts.json`, and `plugins/ca-pi/generated/` is named like a build output but has no generator
I could find. Both stay reviewed.

**`inheritance: true` is load-bearing.** A repo config *replaces* the central
`arbiterForge/coderabbit` config outright without it.

## Verification

Proven before commit, against git's glob semantics (which CodeRabbit's docs say `path_filters` map
onto via `git sparse-checkout`):

- `!plugins/ca/hooks/*.py` excludes **58** hook files
- **55** authored test files under `plugins/ca/hooks/tests/` survive
- **0** of them are caught by the exclusion

Config validated against the live published schema — every key exists, `profile` enum accepted.
Full Python suite green (16 script suites + 1308 unittest tests, exit 0), drift checks green.

## What the review changed

CodeRabbit's own review of this PR surfaced two documentation pages the config's comments were
written without — `configuration/configuration-inheritance` and `reference/glossary`. Both were
verified directly rather than taken at the review's word, and all three corrections held:

1. **Glob semantics are documented, not inferred.** The glossary states `*` matches any filename
   and `**` matches any directory depth. The original comment justified that reading via a
   sparse-checkout mapping and asked for empirical confirmation on the first PR; both were
   unnecessary. **This resolves the main open question — the 55 test files are provably safe.**
2. **Org global overrides are never discarded.** They apply as the final layer after the
   inheritance chain resolves, whatever a repo sets. The old wording overstated the blast radius
   of omitting `inheritance: true`, which affects the central YAML only.
3. **Merge semantics are documented** — objects deep-merge, arrays take the child's items first
   then unique parent items, scalars take the child's value.

## Still not verified

**Whether the central-repo config mechanism works on the free plan.** Undocumented per tier.
`@coderabbitai configuration` on a PR reports the resolved config and the source of each setting,
which is the way to confirm it once `arbiterForge/coderabbit` exists.

## Note on the diff

The `.codearbiter/gate-events.log` append rides along. `git restore --staged` on an audit log is
refused by the H-05 guard (lexical over-breadth, the class tracked in #575), and hunting for a
spelling that evades a gate is not something this lane does. The log is machine-appended session
state.


SUaDtL:
@coderabbitai configuration

PR #685, opened 2026-08-13
docs(spike): preserve findings without spike code

SUaDtL:
## Summary

- define the safe findings-only handoff from a disposable `spike/<slug>` branch to its parent
- require a parent commit through the normal CodeArbiter commit gate before branch deletion
- regenerate Claude, Codex, and Pi command surfaces and cover the required sequence

## Why

A spike may preserve its answer, but not its exploratory implementation. The previous wording required deletion without specifying how the findings file reaches the parent. This makes the transfer explicit while preserving the no-merge and no-spike-code boundaries.

## Verification

- `python .github/scripts/test_spike_findings_contract.py`
- `python tools/build-surface.py --check`
- `python .github/scripts/test_build_surface.py`
- `python .github/scripts/test_host_descriptors.py`
- `python .github/scripts/test_pi_parity.py`
- `python .github/scripts/test_public_pi_docs.py`
- full documented Python commit-gate suite
- `gitleaks protect --no-banner --redact --staged`


SUaDtL:
Superseded by #687. This draft remains preserved for history; #687 carries the same findings-only spike contract from the current main base so CI evaluates only its intended payload.

PR #696, opened 2026-08-16
fix(git-enforce): resolve shared gate markers via marker_root in linked worktrees

### LINKED ISSUE #695: Git backstop must read shared gate markers from linked worktrees
## Problem

In a linked Git worktree, the sanctioned pass producers and the Git commit backstop resolve marker roots differently:

- `migration-pass.py` / `security-pass.py` record content-bound markers at the main checkout's shared `.codearbiter/.markers/` location via `marker_root()`.
- `git-enforce.py pre-commit` correctly resolves the repository being committed for diff and branch checks, but then reads H-09b/H-10b/H-14 marker sets from that linked worktree's private marker directory.

Result: a freshly reviewed migration (or sensitive security diff) can be blocked by the real Git hook even though the sanctioned producer recorded the matching marker. This is a false block, not a missing review.

## Reproduction

1. Create a main checkout with committed `.codearbiter/CONTEXT.md`.
2. Add a linked feature worktree.
3. Stage `db/migrations/0001_init.sql` in the worktree.
4. Run `migration-pass.py` from the worktree with `CLAUDE_PROJECT_DIR` unset.
5. Run `git-enforce.py pre-commit` from that worktree with `CLAUDE_PROJECT_DIR` pinned to the main checkout.

Current behavior before the fix: H-14 blocks because it reads the worktree-local marker set.

The same topology affects H-09b/H-10b after `security-pass.py`.

## Proposed narrow fix

Keep `repo_root()` for the Git operation itself, but use `marker_root()` only for the shared marker file/fingerprint lookups in `git-enforce.py`:

- security marker freshness and marker-set coverage
- migration marker-set coverage

Do not change branch, diff, staged-path, or sensitive-line detection.

## Verified local patch

An isolated patch updates canonical `core/pysrc/git-enforce.py`, then syncs all three shipped hook copies. It adds real linked-worktree regression coverage for both security and migration producer-to-Git-backstop flows.

Evidence:

- New migration regression was RED before the fix: `BLOCKED [H-14]`.
- New security regression was RED before the fix: `BLOCKED [H-09b]`.
- After the narrow lookup change: linked-worktree marker suite 5/5, repository-resolution suite 34/34, marker-freshness suite 8/8, and `sync-core --check` passed.
- The actual SingedTerra pre-commit hook then passed against the staged reviewed migration using the repaired cached plugin.

A broader Git-hook fixture suite requires a POSIX `sh`; on this Windows host its 13 failures were launcher errors due to no `sh`, not assertions in this change.

## Acceptance criteria

- A main-checkout marker created by the sanctioned producer covers the exact staged worktree diff at Git pre-commit time.
- H-09b/H-10b/H-14 remain fail-closed when no matching marker exists.
- Git operation identity remains tied to the worktree executing the commit.
- Canonical core and all shipped host copies remain byte-identical via `sync-core --check`.

snowyukitty:
## Problem
In `core/pysrc/git-enforce.py`, `pre_commit(root)` read gate-pass markers directly from `os.path.join(root, ".codearbiter", ".markers", name)`. Because `.codearbiter/.markers/` is gitignored, linked git worktrees do not share local markers on disk. As a result, gate passes recorded by `security-pass.py` and `migration-pass.py` (which write to `marker_root`) were not visible to pre-commit checks executed in linked worktrees (#695).

## Solution
- Add `_marker_root(root)` helper in `git-enforce.py` that escalates to `git_worktree_main_root(root)` when executing in a linked worktree.
- Update `pre_commit()` to resolve `security-gate-passed` and `migration-gate-passed` markers via `_marker_root(root)` while keeping path-level validation (such as `is_migration_path` and `read_worktree`) anchored to the active worktree checkout.
- Run `tools/sync-core.py` to sync canonical core changes to vendored plugin copies (`plugins/ca/hooks/git-enforce.py`, `plugins/ca-codex/hooks/git-enforce.py`, `plugins/ca-pi/hooks/git-enforce.py`).
- Add unit tests in `plugins/ca/hooks/tests/test_git_enforce_worktree_markers.py` verifying marker root resolution and pre-commit marker reads in both plain repositories and linked worktrees.

Fixes #695

### DIFF
--- a/core/pysrc/git-enforce.py
+++ b/core/pysrc/git-enforce.py
@@ -27,6 +27,7 @@ import sys
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 import hostapi  # noqa: E402 — host seam (ADR-0011)
+from hostapi import git_worktree_main_root  # noqa: E402 — worktree marker-root escalation (#695)
 import _entrylib  # noqa: E402 — shared run() dispatch (jscpd dedup)
 from _gitexec import git_executable  # noqa: E402
 from _hooklib import (  # noqa: E402
@@ -185,6 +186,13 @@ def read_worktree(cwd, rel):
         return None
 
 
+def _marker_root(root):
+    """`root`, escalated to the MAIN checkout when `root` names a LINKED git
+    worktree's checkout (#695) — see `hostapi.git_worktree_main_root`'s
+    docstring."""
+    return git_worktree_main_root(root) or root
+
+
 def _marker_set(root, name):
     try:
         with open(os.path.join(root, ".codearbiter", ".markers", name),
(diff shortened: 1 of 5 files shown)


SUaDtL:
https://github.com/arbiterForge/codeArbiter?tab=contributing-ov-file

PR #732, opened 2026-09-02
feat(adr): bind acceptance to lifecycle evidence

### LINKED ISSUE #674: ADR-0026 is accepted but its mandated registry and CI check were never built
ADR-0026 ("destructive set declared in routing table") is **Accepted**, but the mechanism it mandates does not exist.

## What is missing

- No `## Destructive operations` block exists in `core/surface/includes/routing-table.md` (or any generated copy). Grep matches only the ADR itself and one spec.
- No item-for-item CI checker exists in `.github/scripts/` or `.github/workflows/` comparing ORCHESTRATOR §6's resident copy against that block. `ci.yml` still cites ADR-0022 at the relevant step.
- ADR-0026:64-65 promised a repoint of `test_routing_and_cleanup_surface.py:79` that never landed.

## Why it matters

The only surface carrying the destructive-operations set today is `ORCHESTRATOR.md` §6 (now `arbiter.md`), plus a single test pin. ADR-0026's stated safety property — that the two surfaces are CI-compared so they cannot drift — is not in force. Anyone relying on that guarantee, including work that supersedes part of ADR-0022, has nothing to narrow against.

## Provenance

Found independently by four separate reviewers during planning for #437. Each concluded the registry does not exist; verified by grep and by reading the CI workflow.

### LINKED ISSUE #675: ADRs are immutable by hook but their content is never hashed
H-11 makes `.codearbiter/decisions/` append-only at the filesystem level, but nothing verifies that an accepted ADR's *content* is the content that was accepted.

## The gap

- No hash, checksum, or signature is recorded for any ADR.
- The `/adr` marker that unlocks H-11 is **self-mintable by design** — ADR-0024 rules markers are audit friction, not authorization.
- So an agent (or a person) that mints the marker can edit an accepted ADR in place, and the only evidence is the git diff.

## Why this is not paranoia

This repo already recorded a case of an external agent fabricating an ADR ratification (commit 3902096, corrected via DECISION-0014). The correction worked because someone diffed the governance files by hand. Content hashing would have made that an automated failure rather than a lucky catch.

## Possible shape

Record a content hash in the decision log at acceptance time; add a CI check comparing each accepted ADR's current hash against its recorded one. A legitimate supersession changes state via a *new* ADR, so a hash mismatch on an accepted one is always a defect.

## Provenance

Surfaced during #437 planning while establishing which ADRs could be superseded and how a supersession would be proven. Related: ADR-0024, DECISION-0014.

SUaDtL:
## Summary

This implements ADR-0033 so accepting an architecture decision no longer implies that its obligations are implemented or verified.

- Bind future acceptance to the exact ADR blob, immutable body, and sealed obligation set.
- Record lifecycle transitions in an append-only ledger and derive Implemented or Verified only from complete, fresh evidence.
- Baseline legacy ADRs without inventing historical acceptance or verification evidence.
- Export repository-scoped public guarantees only from Verified obligations.
- Add ADR-0026's destructive-operation registry using ADR-0030's current four-item set and enforce generated Claude, Codex, and Pi parity.
- Advance the affected packages to `ca 2.16.0`, `ca-codex 0.8.0`, and `ca-pi 0.9.0` with synchronized changelogs and generated metadata.

Closes #674
Closes #675

## Why

Accepted ADR text was protected by cooperative write controls but not content-bound to the accepted decision. The repository also lacked a distinct evidence state for implementation and verification. That allowed planned obligations to be mistaken for delivered guarantees. ADR-0026 separately required a destructive-operation registry and parity check that had never been implemented.

## Verification

- ADR lifecycle tests pass 37/37, including explicit-root isolation from ambient repository selectors, and the live lifecycle checker accepts the committed ledger and bindings.
- The destructive registry checker passes all four authoritative surface pairs.
- Hook-guard logic passes 196 assertions with zero failures.
- The full Python hook suite passes 1,457 tests with three intentional skips.
- Pi typecheck, 840-test suite with one intentional skip, build, package, parity, public documentation, and platform-fixture checks pass.
- Shared core, generated surface, host-package, payload-version, badge, plugin-resource, CI-impact, and repository Python checks pass.
- Python syntax, secret scan, provenance rebaseline, and diff checks pass.
- Independent security, auth/crypto, dependency, coverage, and holistic reviews pass on exact head `0fc6336fa6dcf8948e4cebd9f2e8da42e00f7399` with zero findings.

## Coverage note

Python hooks and repository scripts have no numeric coverage tooling. Behavioral coverage comes from focused lifecycle and guard tests plus the complete hook and cross-host suites.

## Architecture and tradeoff

Conflict level 2: ADR-0033 and the repository security controls require truthful lifecycle evidence over the convenience of treating Accepted as delivered. Legacy decisions remain explicitly unsealed, and public claims fail closed until complete repository-scoped evidence reaches Verified.


SUaDtL:
Superseded by the linear-history-safe split: #733 landed the accepted ADR record on main, and #734 carries the lifecycle implementation with its acceptance event bound to that durable mainline commit. This preserves exact source_commit ancestry without force-pushing or changing branch protection.

PR #759, opened 2026-09-07
test(tasks): exercise real taskwrite fidelity

### LINKED ISSUE #628: test fidelity gaps: a vacuous no-op pass in test_ux_conversion.py, a source-grep instead of a real CLI call in test_taskwriter.py
## Finding 1: `test_ux_conversion.py`'s `test_standup_sweep` silently no-ops if both candidate standup files are missing

`.github/scripts/test_ux_conversion.py:158-173`, `test_standup_sweep()`:

```python
for relative in ("core/surface/commands/standup.md",
                 "plugins/ca/commands/standup.md"):
    path = ROOT / relative
    if not path.exists():
        continue
    text = read(path)
    check("taskwrite.py" in text and "archive" in text, ...)
    check("ask about **each one" in text, ...)
```

If **both** candidate paths are missing (renamed, moved, deleted), the loop body never runs, `check()` is never called, and the function returns normally having asserted nothing. `main()`'s harness (`.github/scripts/test_ux_conversion.py:214-229`) accumulates failures in a `_failures` list and reports `"OK: N obligation groups green"` when the list is empty — it has no concept of "this obligation group verified zero things," so a vanished `standup.md` reads identically to a passing test. `ALL_FILES` (the module-level list whose existence `main()` DOES hard-check with a `FATAL` early-exit) does not include either standup.md path, so nothing catches this earlier either.

### Suggested approach

After the loop, assert that at least one candidate path was actually found and checked (fail loudly, not silently, if neither exists) — mirroring the existing `ALL_FILES` FATAL-on-missing pattern for this obligation group specifically.

## Finding 2: `test_taskwriter.py`'s `--desc` CLI-exposure test only source-greps, never actually invokes the CLI

`.github/scripts/test_taskwriter.py:1113-1121`, `test_add_rationale_is_exposed_by_the_taskwrite_cli`:

```python
source = os.path.join(HOOKS, "taskwrite.py")
with open(source, encoding="utf-8") as handle:
    text = handle.read()
self.assertIn('"--desc"', text)
self.assertIn("rationale=args.rationale", text)
```

This proves the two literal substrings exist somewhere in the file — not that `--desc`/`--rationale` actually parse together correctly, that they're wired to the `add` subcommand's dispatch (vs. some other unrelated function containing the same substring by coincidence), or that an end-to-end `taskwrite.py add --desc ... --rationale ...` invocation produces the expected board entry. This is the exact shape the [[dry-run-the-path-not-the-contract]] house pattern warns about: a test proving a mechanism is *present in source* rather than that it *works for the caller's real invocation*.

### Suggested approach

Replace (or supplement) the source-grep with a real subprocess invocation of `taskwrite.py add --desc <x> --rationale <y> ...` against a scratch board fixture, asserting the resulting board text carries both fields — following the pattern this same test file already uses elsewhere for other CLI-surface assertions (worth checking for an existing subprocess-invocation helper in this file to reuse rather than hand-rolling a new one).

## Evidence

See line references above. Found during the #578/#579 findings-triage cluster; cross-ref #578.

SUaDtL:
## Summary

- make the standup sweep obligation fail when neither supported command surface exists
- replace the taskwrite CLI source-grep with a real `add TITLE --desc RATIONALE` subprocess contract
- assert the persisted title, rationale sub-bullet, parsed fields, and queued state

This is a test-only correction. Exact-main defines positional `TITLE` plus `--desc RATIONALE`; it does not define a separate `--rationale` alias, and this PR does not expand that API.

## Test-first evidence

The unchanged external mutation probe failed three assertions against exact base `cf86fe128728c47714ac0632fafeabdc492ed1c7`: both standup surfaces could disappear without failure; the old CLI test stayed green when persisted rationale was discarded; and it stayed green when `--desc` was removed while inert source literals remained. The candidate passes all four probe tests covering six cases.

## Validation

- `python -B .github/scripts/test_taskwriter.py` — 84 tests passed
- `python -B .github/scripts/test_ux_conversion.py` — 8 obligation groups passed
- full required repository command set — 40/40 passed
- full hook discovery — 1,497 tests passed, 3 existing skips
- generated/core/host parity and reference checks passed
- 118 tracked JSON files parsed successfully
- `git diff --check` passed
- gitleaks scanned 23.67 MB with no leaks

The first full-hook attempt retained 14 environment setup errors because Git for Windows `sh` was absent from the PowerShell PATH. The retry added `C:/Program Files/Git/bin` only to that child environment; it then passed with zero errors. The failed receipt remains preserved.

Python hooks and `.github/scripts/*.py` have no numeric coverage tooling. `.codearbiter/tech-stack.md` explicitly states: “There is **no coverage tooling for the Python hooks** (`plugins/*/hooks/*.py`, `.github/scripts/*.py`). No numeric floor exists for those surfaces.” No threshold was changed or waived.

## Review

Independent non-author gpt-6-astra/xhigh coverage review passed with zero findings. The required finding-triage and verdict-aggregator funnel returned `PASS`: 0 BLOCKS, 0 DEFERRABLE, 0 NON_BLOCKING, and 0 incomplete results.

## Hosted provenance follow-up

The first hosted run correctly failed the published-tag immutability check because the automatic release after PR #758 created `v2.17.7`, `ca-codex-v0.9.7`, and `ca-pi-v0.10.8` before their trusted publication receipts were merged. This PR now records only those three receipts from successful release run `34071495941`; it does not change the guard, receipt floor, workflow, checker, tests, or closed legacy epoch.

- all three annotated tag objects peel to exact release commit `cf86fe128728c47714ac0632fafeabdc492ed1c7`
- the strict authenticated audit passes with 136 original-publication receipts and 44 legacy baselines
- focused receipt suites pass 110 tests; the complete 40-command repository gate and 1,497-test hook suite pass
- independent gpt-6-astra/xhigh security review, triage, and aggregation passed with zero findings

The three appended `.codearbiter/gate-events.log` WARN rows were naturally produced by validation across the two commits. The prior 1,191,268 bytes are preserved as an exact prefix; no audit row was rewritten or removed.

Fixes #628

## Summary by CodeRabbit

- **Tests**
  - Replaced source inspection with an end-to-end test covering task creation, saved rationale, and board state.
  - Improved standup command checks to fail when no candidate command files are found.

- **Chores**
  - Recorded three newly published release tags.

### DIFF
--- a/.codearbiter/gate-events.log
+++ b/.codearbiter/gate-events.log
@@ -3866,3 +3866,5 @@ io.open(sys.argv[2],'w',encoding='utf-8',newline='\n').write(open(sys.argv[1],en
 [2026-09-07T00:00:45Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
 [2026-09-07T00:02:29Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
 [2026-09-07T00:06:28Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
+[2026-09-07T00:54:47Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
+[2026-09-07T00:58:45Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
--- a/.codearbiter/gate-events.log
+++ b/.codearbiter/gate-events.log
@@ -3868,3 +3868,4 @@ io.open(sys.argv[2],'w',encoding='utf-8',newline='\n').write(open(sys.argv[1],en
 [2026-09-07T00:06:28Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
 [2026-09-07T00:54:47Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
 [2026-09-07T00:58:45Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
+[2026-09-07T01:31:36Z] WARN host=claude hook=python.exe -m unittest | /mode has been active for over 30 min with no matching overrides.log entry since — confirm the expected audit line landed (CONFIRM-09).
--- a/.github/published-tags.json
+++ b/.github/published-tags.json
@@ -700,6 +700,21 @@
       "commit_sha": "f614620858ac06c4867120987c3bf91b5fa0db10",
       "object_type": "tag",
       "object_sha": "d5a13d272182a26301cec0e9194a2f8cdeb8a5eb"
+    },
+    "v2.17.7": {
+      "commit_sha": "cf86fe128728c47714ac0632fafeabdc492ed1c7",
+      "object_sha": "2acc24e9fd3c9db9b8efef7c24c113347cb2fe17",
+      "object_type": "tag"
+    },
+    "ca-codex-v0.9.7": {
+      "object_type": "tag",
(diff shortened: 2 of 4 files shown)


SUaDtL:
CodeRabbit warning disposition for exact head `b5bd29f6c25717fa055d486ebd44b4100a301365`:

- **Out-of-scope warning: no action.** The receipt append is the narrow, deterministic repair for this PR's exact-head `Published tag immutability` failure after the automatic release triggered by merged PR #758. The PR body now records that causal boundary and the authenticated release evidence. Splitting it out would leave this PR unable to satisfy the required guard; no release logic, checker, receipt floor, or legacy epoch changed.
- **Docstring-coverage warning: no action.** The touched Python functions are regression test methods. This repository's Python validation contract requires behavioral tests and syntax, and defines no docstring threshold for these test surfaces. Adding docstrings would not improve the behavioral proof and would add non-causal churn.

CodeRabbit's exact-head review reports no actionable comments, minimal merge risk, and no review threads. The repository's focused, full, secret, strict provenance, and hosted guards remain authoritative and are green or still running.

